PR #8, opened 2020-05-19
Add Action Syntax - Resize.MinimumPad

patrick-tolosa:


src/actions/resize/minimumPad.js
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @implements IAction
+ */
+class MinimumPad {
+  /**
+   *
+   * @param {number} padWidth
+   * @param {number} padHeight
+   */
+  constructor(padWidth, padHeight) {
```
maoznir, line 12: Why not inherit from some general Resize class?

maoznir: Or implement a resize action interface?

patrick-tolosa: At this moment we don't have all the information of how everything looks like.
I think once we'll have more actions it'll be easier to consolidate.

There's no other Resize action here, so there's nothing to consolidate yet.

patrick-tolosa: Merging for now, we'll discuss reconsolidation once we have more code.

PR #9, opened 2020-05-19
Feature/add action syntax color

patrick-tolosa:
### This PR depends on #8, Can only be reviewed once #8  is merged.

Reviewer - Notice the following:

Color is an action that accepts bgColor as an actionType, as opposed as just accepting the color as an argument, please make sure this is acceptable.


asisayag2:
@patrick-tolosa , this includes two actions - color (which is not a real action) and resize.pad which looks ok

patrick-tolosa:
We'll revisit how 'Color' looks in the context of an argument to other actions.

PR #11, opened 2020-05-25
Feature/add action syntax resize.scale

strausr:


src/actions/resize/scale.js
```diff
@@ -0,0 +1,24 @@
...
+  }
+}
+
+/**
+ * @param {number} scaleWidth
+ * @param {number} scaleHeight
+ * @return {Scale}
+ */
+function scale(scaleWidth, scaleHeight) {
+  return new Scale(scaleWidth, scaleHeight)
```
patrick-tolosa, line 21: Scale also accepts a single argument, maybe the second should default to something?

PR #14, opened 2020-06-03
Feature/add compilation tests

patrick-tolosa:
Notes:

1. A script was added that, given a CODE_GENERATOR_PROJECT, knows to run all the required commands, generate the test file and copy it to the right place. since this involves commands in several languages (bash, node, ruby), some creative solutions were required. 

2. The updateCompliationTests will skip gracefully if specific environment variables are missing, this removes any implicit dependency between this project and the Code Generator project


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "1.0.0-alpha.0",
   "description": "",
   "scripts": {
-    "test": "jest --coverage",
+    "test": "node ./scripts/updateCompliationTests.js && jest --coverage",
```
asisayag2, line 6: If I clone this repo and run `npm run test` will that run? 
I guess I need some Ruby setup on my machine don't I? 
If so, maybe this should be in his own script and only whoever has this set up will choose to run it

patrick-tolosa: No you will not need a local setup.

The script is smart enough to detect if you have the required environment variables, if you don't have them, we just skip this script completely.

This means that Travis, or any other person who has no idea this script exists (and has no ruby) can peacefully run their tests.

If however, a developer has everything set-up, this will generate an updated compilation test and run the tests on that file( as well as the other project files).

PR #20, opened 2020-06-15
Feature/url-creation-and-config

patrick-tolosa:
This is a merge of both PRs - URL and Configuration as the two are dependent on each other.


.eslintrc
```diff
@@ -31,7 +31,6 @@
     "prefer-const": "error",
     "vars-on-top" : "error",
     "no-cond-assign": "error",
-    "no-console" : "warn",
```
patrick-tolosa, line 34: Since we now include console errors in our code, I removed this lint rule to remove noise.

__TESTS__/unit/config/cloudinaryConfig.test.ts
```diff
@@ -0,0 +1,142 @@
...
+import CloudConfig from "../../../src/config/CloudConfig";
+
+describe('Tests for CloudinaryConfiguration', () => {
+  it('Creates a CloudinaryConfig with defaults', () => {
+    const conf = new CloudinaryConfig({});
+
+    expect(conf.account).toEqual({});
+    expect(conf.cloud).toEqual({});
+    expect(conf.url).toEqual({});
+  });
```
asisayag2, line 13: won't you get an error if you don't provide a cloudName?

__TESTS__/unit/urlConfig.test.ts
```diff
@@ -0,0 +1,74 @@
...
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/my_image');
+  });
+  it ('Generates a URL with transforamtions', () => {
+    const url = new TransformableImage()
+      .setConfig(DEMO_CONFIG)
+      .resize(Resize.fill(100, 100))
+      .setPublicID('sample')
+      .describeAsset({
+        // has sane defaults for image/upload
+      })
```
asisayag2, line 27: why do we need that?

__TESTS__/unit/urlConfig.test.ts
```diff
@@ -0,0 +1,74 @@
...
+        cloudName: 'MY_CLOUD_NAME'
+      }
+    });
+
+    /**
+     * Use the "global" instance
+     */
+    const url = myInstance.image('sample')
+      .resize(Resize.fill(100, 100))
+      .describeAsset({ // In PHP, thi
```
asisayag2, line 66: Comment is cut

asisayag2: Also, why not `assetDescriptor`? not sure `describeAsset` means much

src/actions/resize/crop.ts
```diff
@@ -1,17 +1,29 @@
 import '../../interfaces/IAction';
 import Action from "../Action";
+import {IAction} from "../../interfaces/IAction";
 
 /**
  * @implements IAction
  */
 class Crop extends Action implements IAction {
```
asisayag2, line 8: This is only initial, right? 
cause lots of things are missing - gravity, ar, etc.

src/config/BaseConfig.ts
```diff
@@ -0,0 +1,28 @@
...
+import {INVALID_TYPE_MESSAGE} from "../constants";
+
+class Config {
+  /**
+   * Type guard for user config input
+   * @param {any} object User provided input of unknown shape
+   * @param {string[]} validKeys Array of valid configuration keys
+   * @return {boolean} Whether an object is a valid configuration object
+   */
+  isValidConfig(object:unknown, validKeys:string[]): object is ICloudConfig {
```
asisayag2, line 11: why `ICloudConfig`? not supposed to be `ICloudinaryConfig`?

src/config/CloudConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+import Config from "./BaseConfig";
+
+
+
+class CloudConfig extends Config implements ICloudConfig {
+  cloudName?: string;
+  apiKey?: string;
+  apiSecret?: string;
+
+  [other:string]: unknown;
```
asisayag2, line 12: Didnt we decided to drop that?

src/config/CloudConfig.ts
```diff
@@ -0,0 +1,37 @@
...
+   * @param {ICloudConfig} cloudConfig
+   */
+  constructor(cloudConfig: ICloudConfig | unknown) {
+    super();
+    if (this.isCloudConfigTypeGuard(cloudConfig)) {
+      Object.assign(this, cloudConfig);
+    } else {
+      console.error(INVALID_TYPE_MESSAGE);
+    }
+  }
```
asisayag2, line 24: Just to make sure I understand the logics - if I have a typo in one of the keys, the entire thing gets ignored? 
Why not ignoring just the invalid keys? 

I feel like we're doing a lot of work the TS supposed to do... which brings me to the question of "why unknown"? if we drop that, TS will handle all validations for us, no?

src/interfaces/Config/ICloudinaryConfigurations.ts
```diff
@@ -0,0 +1,12 @@
+import IAccountConfig from "./IAccountConfig";
+import ICloudinaryAssetConfigurations from "./ICloudinaryAssetConfigurations";
+import ITagConfig from "./ITagConfig";
+import IApiConfig from "./IApiConfig";
+
+interface ICloudinaryConfigurations extends ICloudinaryAssetConfigurations{
+  account?: IAccountConfig
```
asisayag2, line 7: We said we will have this totally separated from the rest

src/interfaces/Config/ICloudinaryConfigurations.ts
```diff
@@ -0,0 +1,12 @@
...
+  tag?: ITagConfig
+  api?: IApiConfig
```
asisayag2, line 9: Where are `URLConfig` and `CloudConfig`?

src/interfaces/Config/ITagConfig.ts
```diff
@@ -0,0 +1,16 @@
+interface ITagConfig {
+  responsive?: boolean;
+  responsiveClass?: string;
+  responsiveWidth?: boolean;
+  responsivePlaceholder?: string;
```
asisayag2, line 5: We decided to have IResponsiveConfig as a child of ITagConfig

src/interfaces/Config/IURLConfig.ts
```diff
@@ -0,0 +1,15 @@
+interface IURLConfig {
+  cdnSubdomain?: boolean;
```
asisayag2, line 2: missing `analytics`

src/url/cloudinaryURL.ts
```diff
@@ -7,9 +9,60 @@ import Transformation from '../transformation/Transformation';
  * @param {Object} descriptor
  * @param {Transformation} transformation
  */
-function createCloudinaryURL(config: unknown, descriptor: unknown, transformation: unknown) {
-  // Currently not implemented
-  return '';
+function createCloudinaryURL(config: CloudinaryConfig, descriptor?: IDescriptor, transformation?: Transformation) {
```
asisayag2, line 12: is that a public interface?

__TESTS__/unit/config/cloudinaryConfig.test.ts
patrick-tolosa: This is the main "CloudinaryConfig" instance, 

It's expected to have a certain shape (accoumt, cloud, url), but we don't expect anything to be mandatory.

__TESTS__/unit/urlConfig.test.ts
patrick-tolosa: This is an optional function that allows us to add additional asset information (such as type, resourceType etc.)

You don't have to use it, this is strictly for showing purposes that it exists with some defaults

__TESTS__/unit/urlConfig.test.ts
patrick-tolosa: I think describeAsset means much more than assetDescriptor. (First of all, one is verb and the other is a noun, but also assetDescriptor is an internal term for Cloudinary, while describeAsset could be more generic.

No doubt that "describeAsset" is not the best name,  but assetDescriptor is exactly the same only with reversed words. 

I do want to find a better name though.

src/actions/resize/crop.ts
patrick-tolosa: Yes, we'll be working through the full compilation tests anyway, this is a proof of concept.

src/config/BaseConfig.ts
patrick-tolosa: ICloudinaryConfig is the main Configuration object(that includes cloud, account, url, api etc.)

ICloudConfig is the shape for the CloudConfiguration.

src/interfaces/Config/ICloudinaryConfigurations.ts
patrick-tolosa: I did not know that this is what was agreed.
If we're sure we can remove the AccountAPI code from this PR.

src/interfaces/Config/ICloudinaryConfigurations.ts
patrick-tolosa: Inside ICloudinaryAssetConfigurations

(notice this is an extension)

src/interfaces/Config/ITagConfig.ts
patrick-tolosa: I wasn't aware of that

src/url/cloudinaryURL.ts
patrick-tolosa: In theory it could be, it's currently used through the transformableImage class

```  
toURL(): string {
    return createCloudinaryURL(this.config, Object.assign({
      resourceType: 'image',
      type: 'upload'
    }, this.asset), this);
  }
```

PR #26, opened 2020-06-29
Add overlay action

patrick-tolosa:
Notes.

1. This PR includes a fix for joining actions with '/' 
2. Note the difference between PHP and this PR, for size reasons we can't use the same approach.
3. This is not aligned to the code generator, which will require changes.
4. this is not the complete overlay implementation, but merely a POC to decide if this is how we want to proceed

### DIFF
--- a/__TESTS__/unit/actions/Overlay.test.ts
+++ b/__TESTS__/unit/actions/Overlay.test.ts
@@ -1,0 +1,33 @@
+import {fill} from "../../../src/actions/resize/Resize";
+import {overlayLayer} from "../../../src/actions/layers/Layers";
+import {imageSource} from "../../../src/actions/layers/Sources";
+import TransformableImage from "../../../src/transformation/TransformableImage";
+
+describe('Tests for overlay actions', () => {
+  it('Parses an overlay with an imageSource', () => {
+    const tImage = new TransformableImage('sample');
+    tImage
+      .setConfig({
+        cloud: {
+          cloudName: 'demo'
+        }
+      })
+      .resize(fill(1000, 1000))
+      .overlay(
+        overlayLayer(
+          imageSource('sample')
+            .resize(fill(500, 500))
+            .overlay(
+              overlayLayer(
+                imageSource('sample')
+                  .resize(fill(200, 200)
+                  )
+              )
+            )
+        )
+      );
+
+    expect(tImage.toURL())
+      .toBe('http://res.cloudinary.com/demo/image/upload/c_fill,w_1000,h_1000/l_sample/c_fill,w_500,h_500/l_sample/c_fill,w_200,h_200/fl_layer_apply/fl_layer_apply/sample');
+  });
+});
--- a/__TESTS__/unit/transformation/TransformableImage.test.ts
+++ b/__TESTS__/unit/transformation/TransformableImage.test.ts
@@ -55,6 +55,6 @@ describe('Tests for TransformableImage', () => {
         }
       });
 
-    expect(tImage.toString()).toBe('aabb');
+    expect(tImage.toString()).toBe('aa/bb');
   });
 });
--- a/src/actions/layers/Sources.ts
+++ b/src/actions/layers/Sources.ts
@@ -1,0 +1,15 @@
+import TransformableImage from "../../transformation/TransformableImage";
+
+export class ImageSource extends TransformableImage {
+}
+
+/**
+ * @param publicID
+ */
+function imageSource(publicID: string) {
+  return new ImageSource(publicID);
+}
+
+export {
+  imageSource
+};
--- a/src/transformation/Transformation.ts
+++ b/src/transformation/Transformation.ts
@@ -19,13 +19,9 @@ class Transformation {
    * @return {string}
    */
   toString(): string {
-    let str = '';
-
-    this.actions.forEach((action) => {
-      str += action.toString();
-    });
-
-    return str;
+    return this.actions.map((action) => {
+      return action.toString();
+    }).join('/');
   }
 }
 
--- a/src/actions/layers/ILayerAction.ts
+++ b/src/actions/layers/ILayerAction.ts
@@ -1,0 +1,5 @@
+import {IAction} from "../../interfaces/IAction";
+
+export interface ILayerAction extends IAction{
+
+}
--- a/src/actions/layers/Layers.ts
+++ b/src/actions/layers/Layers.ts
@@ -1,0 +1,31 @@
+import {ILayerAction} from "./ILayerAction";
+import Action from "../Action";
+import {ImageSource} from "./Sources";
+
+class Layer extends Action implements ILayerAction {
+  source: ImageSource;
+  // TODO this needs to accept other types of sources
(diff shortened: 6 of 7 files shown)


src/transformation/Transformation.ts
```diff
@@ -19,13 +19,9 @@ class Transformation {
...
    */
   toString(): string {
-    let str = '';
-
-    this.actions.forEach((action) => {
-      str += action.toString();
-    });
-
-    return str;
+    return this.actions.map((action) => {
```
patrick-tolosa, line 22: A fix for concatinating actions with '/'

__TESTS__/unit/transformation/TransformableImage.test.ts
```diff
@@ -55,6 +54,6 @@ describe('Tests for TransformableImage', () => {
         }
       });
 
-    expect(tImage.toString()).toBe('aabb');
+    expect(tImage.toString()).toBe('aa/bb');
```
patrick-tolosa, line 57: A fix for concatinating actions with '/'

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -0,0 +1,30 @@
...
+      .setConfig({
+        cloud: {
+          cloudName:'demo'
+        }
+      })
+      .resize(fill(200, 200))
+      .quality(best())
+      .overlay(
+        // imageLayerAction is what registers on the transformation(which has toString() method)
+        // imageSource is the parameter for imageLayerAction
```
patrick-tolosa, line 20: Comment needs to be fixed 
it's not imageSource but rather imageLayer

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        imageLayer( imageSource('sample')
+          .resize(fill(50, 50))
+          .quality(auto())
+        )
+      );
+
+    expect(tImage.toURL())
+      .toBe('http://res.cloudinary.com/demo/image/upload/c_fill,w_200,h_200/q_auto:best/l_sample,c_fill,w_50,h_50,q_auto/sample');
```
maoznir, line 28: I know this isn't aligned with the spec but still i'd like to point our that eventually here you need to add fl_layer_apply at the end of the layer and move the actions to different components

src/actions/layers/overlay.ts
```diff
@@ -0,0 +1,28 @@
+import {IOverlayAction} from "./IOverlayAction";
+import Action from "../Action";
+import {ImageLayer} from "./Sources";
+
+
+
+class Overlay extends Action implements IOverlayAction {
+  imageLayer: ImageLayer;
```
maoznir, line 8: Should have a general source or layer, to fit video/image/text

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        }
+      })
+      .resize(fill(200, 200))
+      .quality(best())
+      .overlay(
+        // imageLayerAction is what registers on the transformation(which has toString() method)
+        // imageSource is the parameter for imageLayerAction
+        imageLayer( imageSource('sample')
+          .resize(fill(50, 50))
+          .quality(auto())
```
nitzanj, line 23: That is the least important thing in this review, however it's still an inaccuracy - being transformable, layers to support format/quality actions (since they support all actions), however it's quite meaningless to use these actions on a layer. So while not really wanting to block it in the code (it's not even blocked on the backend), we do try to have realistic use-cases in our public tests.

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -0,0 +1,30 @@
...
+        cloud: {
+          cloudName:'demo'
+        }
+      })
+      .resize(fill(200, 200))
+      .quality(best())
+      .overlay(
+        // imageLayerAction is what registers on the transformation(which has toString() method)
+        // imageSource is the parameter for imageLayerAction
+        imageLayer( imageSource('sample')
```
nitzanj, line 21: I'm not sure how you add the rest of the layer-action components here (mainly position and blendmode). I'm not saying the PoC needs the full implementation, but we do need to verify it's doable and ends up with a decent API once those are added.

src/actions/layers/Sources.ts
```diff
@@ -0,0 +1,23 @@
+import TransformableImage from "../../transformation/TransformableImage";
+
+export class ImageLayer extends TransformableImage {
+  // TODO So an ImageLayer needs to stringify itself differently
+  // instead of joining with '/' we join with ','
+  // wil be reused probably
+  toString(): string {
```
nitzanj, line 7: This is a bit trickier than that. A layer action - and all it's derivatives (cutter, style_transfer, etc.) - is built using more than one component, unlike all other actions (component = anything between slashes in the transformation).
We should probably schedule a short session detailing this, however it'll end up a lot more complicated than this join. 

Another issue with our complicated type-hierarchy is that a lot of the layer behaviour is shraed among all layer types, however they can't have a shared super-class - since each one has a super class already (e.g. this class has `TransformableImage`, and the video layer will have another). That's why in PHP there are traits everywhere, and that's why in Kotlin it's just completely different altogether.

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -0,0 +1,33 @@
...
+    const tImage = new TransformableImage('sample');
+    tImage
+      .setConfig({
+        cloud: {
+          cloudName: 'demo'
+        }
+      })
+      .resize(fill(1000, 1000))
+      .overlay(
+        addImageLayer(
```
maoznir, line 17: I suggest to keep the signature of (source/layer, position/gravity, blendMode)

patrick-tolosa: In JS we need an extra layer: `addImageLayer` which is the actual action.

`addImageLayer` will accept (probably) three arguments:

Source (`imageSource`), `Position` and `blendMode`

this PR doesn't contrast that, but just implements the first part of the overlay

src/actions/layers/Sources.ts
```diff
@@ -0,0 +1,15 @@
+import TransformableImage from "../../transformation/TransformableImage";
+
+export class ImageLayer extends TransformableImage {
```
patrick-tolosa, line 3: typo, should be ImageSource

src/actions/layers/overlay.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+}
+
+/**
+ * @return {Overlay}
+ */
+function addImageLayer(imageSource: ImageLayer, a?:any, b?:any): Overlay {
+  return new Overlay(imageSource);
+}
+
```
patrick-tolosa, line 25: Worth refactoring to something like this
`.overlay(overlayLayer(
			  Source.image('cloudinary_icon')
          .adjust(Adjust.opacity(50))
          .resize(Resize.scale(60))
        )
      );`

PR #27, opened 2020-06-30
Add Param, Flag, ParamValue

maoznir:


__TESTS__/unit/Action/Action.test.ts
```diff
@@ -0,0 +1,100 @@
...
+      .addParam(new Param('l', 'sample'))
+      .addFlag(new Flag('layer_apply'));
+
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .addAction(action)
+      .setPublicID('sample')
+      .toURL();
+
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/l_sample,fl_layer_apply/sample');
```
patrick-tolosa, line 84: I think layers will behave differently.
but this test does prove the point that flags work as a Param though.

PR #28, opened 2020-06-30
Sort params on an action before toString, use Params inside Crop action

patrick-tolosa:


src/actions/Action.ts
```diff
@@ -6,18 +6,30 @@ class Action implements IAction {
   // We're using map, to overwrite existing keys. for example:
   // addParam(w_100).addParam(w_200) should result in w_200. and not w_100,w_200
   params: Map<string, Param> = new Map();
-  delimiter = ',' // {param}{delimiter}{param} for example: `${'w_100'}${','}${'c_fill'}`
+  delimiter = ','; // {param}{delimiter}{param} for example: `${'w_100'}${','}${'c_fill'}`
 
-  toString() {
-    return Array.from(this.params.values()).join(this.delimiter);
+  getSortedParams(): Param[] {
```
maoznir, line 11: Future improvement: this and any "util" function should be static or imported from utils

src/actions/resize/crop.ts
```diff
@@ -1,28 +1,21 @@
 import Action from "../Action";
```
maoznir, line 1: I don't think there should be a crop action, crop_mode is a param of Resize

patrick-tolosa: At this point, I just refactored the existing crop to use Params,
However, I can see how in the very near future we'll join all Resize into a single class.

We'll do it on a separate PR as it'll change a few files.

src/actions/Action.ts
patrick-tolosa: I'm even tempted to somehow have `params.sort()`  (we currently use a map, but maybe we could use our own wrapper over it)

maoznir: yeah good idea, lets merge and refactor

PR #29, opened 2020-07-01
added signature formatting

strausr:


src/url/cloudinaryURL.ts
```diff
@@ -17,11 +17,12 @@ function createCloudinaryURL(config: CloudinaryConfig, descriptor?: IDescriptor,
   const prefix = getUrlPrefix(config.cloud.cloudName);
   const assetType = handleAssetType(descriptor);
   const storageType = handleStorageType(descriptor);
+  const signature = descriptor.signature;
   const transformationString = transformation ? transformation.toString() : '';
   const version = getUrlVersion(config.url, descriptor);
   const publicID = descriptor.publicID;
 
-  const url = [prefix, assetType, storageType, transformationString, version, publicID]
+  const url = [prefix, assetType, storageType, signature, transformationString, version, publicID]
```
patrick-tolosa, line 25: Shouldn't we check if no signature was provided?
Can Signature be undefined? 
What happens in that case?

strausr: I could add a check but do you think it's necessary? The user either passes it in or not. You can see in all the other tests on cloudinaryURL.ts that no signature is being passed in

PR #30, opened 2020-07-02
Add Rotate action

maoznir:


src/actions/Action.ts
```diff
@@ -10,13 +10,9 @@ class Action implements IAction {
...
-      if (param1.key < param2.key) {
-        return -1;
-      } else if (param1.key > param2.key) {
-        return 1;
-      } else {
-        return 0;
-      }
+      // No need to check for equality because params is a map.
+      // Also, returning 0 will be unreachable code that will break code coverage.
+      return (param1.key > param2.key) ? 1 : -1;
```
patrick-tolosa, line 15: The return 0 is non reacahable, but it's part of the spec of sort (It can be confusing reading this sort without an equality condition).

I don't think this function should assume the type of params(now it's a map, but it could change)

If we want we can strictly check that params is a map by checking that `params` is instanceof map.
I agree that the equality condition is never reached in the current implementation.

src/actions/rotate/IRotateAction.ts
```diff
@@ -0,0 +1,7 @@
+import {IAction} from "../../interfaces/IAction";
+
+export interface IRotateAction extends IAction{
+  angle(angle: number): IRotateAction;
```
patrick-tolosa, line 4: Are these functions the interface of Action? (Do all Actions implement these functions?)

We probably need an IRotateAction interface

src/actions/rotate/Rotate.ts
```diff
@@ -0,0 +1,42 @@
...
+  constructor(angle?: number) {
+    super();
+    this.addParam(new Param('a', angle));
+  }
+
+  /**
+   * @return {Rotate}
+   */
+  horizontalFlip(): Rotate {
+    this.params.get('a').addValue('hflip');
```
patrick-tolosa, line 18: maybe rename to getByKey? not critical though

patrick-tolosa: This feels a bit strange,
For addParam, we have a proper method.
however to reach down into the params itself to get a key.

Can we add a function to getParamByKey? do you think it makes sense?

src/actions/Action.ts
maoznir: I suggest that if/when this becomes a util, it will be called sortMapByKeys or something similar. I've added the comments that explain this, so that any future dev can easily understand and make any change.

src/actions/rotate/Rotate.ts
maoznir: done

src/parameters/ParamValue.ts
```diff
@@ -7,17 +7,22 @@ class ParamValue {
...
   addValue(value: unknown) {
+    //Append values or array of values
     if (Array.isArray(value)) {
       this.values = this.values.concat(value);
     } else {
       this.values.push(value);
     }
 
+    //remove empty values
+    this.values = this.values.filter((v) => v != null);
```
patrick-tolosa, line 24: What do you think of 
this.values.filter((v) => v); (remove false, null, undefined, '' any aother falsy element)

src/actions/Action.ts
patrick-tolosa: My only concern is that comments lie (eventually, if the implementation will change)
I still prefer to have either a strict check or just include the equality operation.

It shouldn't break the code coverage, but if we're really concerned we could omit that logical branch from code coverage 

https://github.com/gotwarlost/istanbul/blob/master/ignoring-code-for-coverage.md

src/actions/rotate/Rotate.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  horizontalFlip(): Rotate {
+    return this.addValue(HORIZONTAL_FLIP);
+  }
+
+  /**
+   * @return {Rotate}
+   */
+  verticalFlip(): Rotate {
+    return this.addValue(VERTICAL_FLIP);
```
patrick-tolosa, line 28: Add value - to which Param though?

An action can have multiple params

src/actions/rotate/Rotate.ts
```diff
@@ -0,0 +1,47 @@
+import horizontalFlip from "./horizontalFlip";
+import verticalFlip from "./verticalFlip";
+import angle from "./angle";
+import Action from "../Action";
+import {IAction} from "../../interfaces/IAction";
+import Param from "../../parameters/Param";
+const HORIZONTAL_FLIP = 'hflip';
+const VERTICAL_FLIP = 'vflip';
+const PARAM_KEY = 'a';
```
patrick-tolosa, line 9: PARAM_KEY is a param-level variable, a single action can have multiple params, so I'm not sure this is the right approach

maoznir: Rotate is a single param action, future refactor can group those under SingleParamAction

src/actions/rotate/Rotate.ts
maoznir: it's private because it's a single param action

src/parameters/ParamValue.ts
maoznir: agree

src/actions/Action.ts
maoznir: Moved to sortMapByKey

src/actions/rotate/Rotate.ts
patrick-tolosa: This feels like a detail though, especially as this is completely out of our control whether something is a single or multi param.

This adds constraints we don't necessarily want and that we might eventually regret. 

Once this becomes a multiParam action, a lot needs to change in this Action.

maoznir: please suggest different solution

src/utils/Utils.ts
```diff
@@ -0,0 +1,5 @@
+import {sortMapByKey} from './dataStructureUtils';
+
+export {
```
patrick-tolosa, line 3: Do we really need the extra hop? 
Let's just import directly from the file we want 

(`import sortMapByKey from utils/dataStructureUtils)`
Typescript makes these imports really simple.

src/utils/dataStructureUtils.ts
```diff
@@ -0,0 +1,17 @@
+import {IKeyObject} from "../interfaces/IKeyObject";
+
+/**
+ * Sort a map of IKeyObject by key
+ * @param {<string, IKeyObject>)} map to sort
+ * @Return sorted map
+ */
+function sortMapByKey(map: Map<string, IKeyObject>) {
+  return (Array.from(map.values()) as IKeyObject[]).sort((obj1, obj2) => {
```
patrick-tolosa, line 9: Looking at IKeyObject (the interface), this can probably be simplified:

`function sortMapByKey(map: Map<string, {key: unknown}>) {`

and no need for the `as...` syntax since we know the types
`return Array.from(map.values()).sort((obj1, obj2) => {`

The reason I suggest these changes is because the code written in this PR won't provide any typesafey, since the key itself is optional and any other key can be passed, we don't benefit at all from this type.

Let's keep it minimal, and if we'll need more references to something with a key we can export it out

src/actions/rotate/Rotate.ts
patrick-tolosa: We'll PR it once this is merged as this is an internal detail that doesn't affect the feature itself.

src/interfaces/IKeyObject.ts
```diff
@@ -0,0 +1,5 @@
+export interface IKeyObject {
+  key?: string;
```
patrick-tolosa, line 2: See my comments below in dataStructureUtils.ts

PR #31, opened 2020-07-02
Added border

strausr:


src/actions/Action.ts
```diff
@@ -7,6 +7,7 @@ class Action implements IAction {
   // addParam(w_100).addParam(w_200) should result in w_200. and not w_100,w_200
   params: Map<string, Param> = new Map();
   delimiter = ','; // {param}{delimiter}{param} for example: `${'w_100'}${','}${'c_fill'}`
+  prepareParam() {}
```
maoznir, line 10: I don't think prepareParam should be in Action, but we can refactor this later.
@patrick-tolosa what do you think

patrick-tolosa: This is the equivalent of `build` in the other SDKs.

The problem we have here is that we can create the final string only when we have all the information, regardless of the order of insertion.

Notice how `prepareParam` contains action-specific business logic, and it's not something generic. 

We could have it as an external function `prepareBorderParam` but I don't see much merit in spreading the logic to more files

src/actions/Action.ts
```diff
@@ -32,8 +34,9 @@ class Action implements IAction {
   addFlag(flag: Flag): this {
     const existingFlag = this.params.get('fl_');
     if (existingFlag){
-      existingFlag.addValue(flag.paramValue);
+      existingFlag.addValue(flag.paramValue.setDelimiter('.'));
```
patrick-tolosa, line 37: Seems like we do this regardless of the if branches, can this be moved to the top?

`flag.paramValue.setDelimiter('.')`

PR #33, opened 2020-07-02
Refactor how we extend the paramValue in rotate

patrick-tolosa:


src/actions/rotate/Rotate.ts
```diff
@@ -4,41 +4,40 @@ import angle from "./angle";
 import Action from "../Action";
 import {IAction} from "../../interfaces/IAction";
 import Param from "../../parameters/Param";
+import ParamValue from "../../parameters/ParamValue";
```
maoznir, line 7: This looks much worse to me, Action should not contain a paramValue.
Also, action already contains a params map.

patrick-tolosa:
Closing this in favour of  https://github.com/cloudinary/cloudinary-js-base/pull/34

src/actions/rotate/Rotate.ts
patrick-tolosa: This is how I feel about an action containing a PARAM_KEY

An action should not be aware that it has one or many params.
However, **we are going to need that, either knowing a key or knowing about a value**

PR #34, opened 2020-07-02
Add mergeParam to Action and use it in Rotate

maoznir:


src/actions/rotate/Rotate.ts
```diff
@@ -18,28 +18,21 @@ class Rotate extends Action implements IAction {
    * @return {Rotate}
    */
   horizontalFlip(): Rotate {
-    return this.addValue(HORIZONTAL_FLIP);
+    return this.mergeParam(new Param(PARAM_KEY, HORIZONTAL_FLIP));
```
patrick-tolosa, line 21: Sounds really good,
I would only change the parameters to be "key" and "value" instead of a Param Instance.

With this approach, we're taking the key/value anyway from the Param instance. 
What do you think?

maoznir: I think we should leave it as param so be similar to addParam

patrick-tolosa: when you merge two objects, there's a very specific behaviour, this gives me the sensation we're mergin two `new Params(...)`, when we're not, we're merging their values.

If we accept a whole param as input, the name should reflect it better(we're merging two objects, but rather adding a value)

in any case, this isn't exactly a classic merge.

@strausr  - What do you think?

We'd like to ability to extend/merge a param with more values
`this.mergeParam(new Param(PARAM_KEY, VALUE));` will find an already existing  param with a key `PARAM_KEY`, and then call `.addValue(VALUE)`. 

What do you think of the name and arguments? do you have any suggestions?

strausr: I tend to agree with you. Looking at this without an explanation looks like we're trying to merge two new params. I think we could either change the parameters to make it clearer or change the name.. how about mergeParamWithValue or mergeParamValue

maoznir: ok i'll change it to addValueToParams(paramKey, paramValue)

PR #35, opened 2020-07-03
Add Variable action

maoznir:


src/utils/dataStructureUtils.ts
```diff
@@ -1,12 +1,31 @@
...
+  return array.map((v) => v[1]);
+}
+
+/**
+ * Checks if `value` is classified as a `String` primitive or object.
+ * Based on lodash/isString
+ * @param {*} value The value to check.
+ * @returns {boolean} `true` if `value` is a string, else `false`.
+ */
+function isString(value: unknown) {
```
patrick-tolosa, line 18: I think this is an overkill to what we need, 

We could probably do something simpler 
`if (typeof myVar === 'string' || myVar instanceof String)`

We basically want to support `'foo'` and `new String('foo')` - I'm not even sure we should support `new String('foo')` to be honest.


(Taken from this stack overflow https://stackoverflow.com/questions/4059147/check-if-a-variable-is-a-string-in-javascript)

src/actions/variable/Variable.ts
```diff
@@ -0,0 +1,29 @@
...
+
+// TODO: when spec for $award_ctx:!First! is ready, implement it
+// TODO: when spec for $award_ref:!First! is ready, implement it
+// TODO: when spec for to_i & to_f is ready, implement it
+
+class Variable extends Action implements IVariableAction {
+  constructor(key: string, value: number | string) {
+    super();
+    const paramKey = `$${key}`;
+    const paramValue = isString(value) ? `!${value}` : value;
```
patrick-tolosa: Possible values are only strings or numbers, 
It might be easier to check if something is a number?

__TESTS__/unit/actions/Variable.test.ts
```diff
@@ -0,0 +1,62 @@
...
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/$a_30/sample');
+  });
+  it('Creates a cloudinaryURL with variable', () => {
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .variable(variable('a', 'bbb'))
+      .setPublicID('sample')
+      .toURL();
+
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/$a_!bbb/sample');
```
patrick-tolosa: These don't work in the browser, probably missing a closing !

__TESTS__/unit/actions/Variable.test.ts
```diff
@@ -0,0 +1,62 @@
...
+      .variable(variable('a', 30))
+      .setPublicID('sample')
+      .toURL();
+
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/$a_30/sample');
+  });
+  it('Creates a cloudinaryURL with variable', () => {
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .variable(variable('a', 'bbb'))
```
strausr, line 38: looks like a good start but there are some cases we are not taking care of. If this is something that we want to add to in the future then we should track the cases that we still did not implement. 
For example,
`.variable(variable('list', ['b', 'b', 'b'))` should yield $list_!b:b:b!

Maybe we could look at the spec and track what was implemented and what was not?

patrick-tolosa: Let's add these use cases as skipped( or empty) tests, so we know to go back to them.

src/utils/dataStructureUtils.ts
maoznir: ok

__TESTS__/unit/actions/Variable.test.ts
maoznir: I've added a TODO in Variable. we'll add the tests when the spec is ready

src/actions/rotate/Rotate.ts
```diff
@@ -2,13 +2,13 @@ import horizontalFlip from "./horizontalFlip";
...
 import Action from "../Action";
-import {IAction} from "../../interfaces/IAction";
 import Param from "../../parameters/Param";
+import {IRotateAction} from "./IRotateAction";
 const HORIZONTAL_FLIP = 'hflip';
 const VERTICAL_FLIP = 'vflip';
 const PARAM_KEY = 'a';
 
-class Rotate extends Action implements IAction {
+class Rotate extends Action implements IRotateAction {
```
patrick-tolosa, line 11: Seems unrelated to Variable

maoznir: Reverted this

PR #36, opened 2020-07-05
Align exports so that Rotate.angle is a function

patrick-tolosa:


src/actions/rotate/Rotate.ts
```diff
@@ -17,31 +17,31 @@ class Rotate extends Action implements IAction {
...
       this.params.get(PARAM_KEY).addValue(value);
     }
     return this;
   }
 }
 
-export {angle, horizontalFlip, verticalFlip};
-export default Rotate;
+export {angle, horizontalFlip, verticalFlip, RotateAction};
+export default {angle, horizontalFlip, verticalFlip, RotateAction};
```
maoznir, line 47: Why do we need to export default anything but RotateAction?

patrick-tolosa: This will allow developers to use `import Rotate from 'actions/rotate' 

And then use it as a namespace
`Rotate.angle()` or `Rotate.verticalFlip()` 

The named exports are then used for those who want an even smaller bundle size 
`import {angle} from 'actions/rotate'`

maoznir: But this will be enough no?
export {angle, horizontalFlip, verticalFlip};
export default RotateAction;

patrick-tolosa: RotateAction is a class, and it has no static members (RotateAction.angle), so it wouldn't be exactly the same

PR #39, opened 2020-07-08
Add better-docs template to the repo

patrick-tolosa:


package-lock.json
```diff
@@ -907,6 +907,16 @@
         "@babel/helper-plugin-utils": "^7.8.3"
```
strausr, line 907: We can delete this no?

patrick-tolosa: Part of package lock, related to a bigger question of do we want package.lock or not.

PR #42, opened 2020-07-08
Add action- Color

strausr:


__TESTS__/unit/actions/Color.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  it('Ensure namespace is correctly populated', () => {
+    expect(Color.rgb).toEqual(rgb);
+    expect(Color.namedColor).toEqual(namedColor);
+  });
+
+  it('Ensures rgb is accepted as an action to TransformableImage', () => {
+    const tImage = new TransformableImage();
+    // Ensures it compiles and doesn't throw
+    expect(
+      tImage.color(rgb(Colors.AZURE2))
```
patrick-tolosa, line 23: Please add a test that checks that AZURE2 can be used alone:

`import {AZURE2} from '.../Colors`

__TESTS__/unit/actions/Color.test.ts
```diff
@@ -0,0 +1,64 @@
+import TransformableImage from '../../../src/transformation/TransformableImage';
+import CloudinaryConfig from "../../../src/config/CloudinaryConfig";
+import {Color} from "../../../src/actions/Actions";
+import Colors from '../../../src/actions/color/Colors';
```
patrick-tolosa, line 4: This will not work for our users,

`import Colors from '@cloudinary/base/actions/color/Colors'` ill fail as it's not a real path.

Perhaps, to avoid confusion, /Actions/ should only contain Actions?
We could have a separate place for Constants
Aside from fixing it now, do you have any idea on how we can make sure this doesn't happen again?

__TESTS__/unit/actions/Color.test.ts
strausr: Do you mean this import AZURE2 = Colors.AZURE2; ?

__TESTS__/unit/actions/Color.test.ts
strausr: The only thing I can think of is having a consumer as part of the workspace.. but it will require upkeeping two separate test cases. Also, we can test/search for a tool such as this- https://www.npmjs.com/package/pack-n-play this will enable us to verify that our library can be used by dependent code (the d.ts we're feeding it but this might not be enough)

__TESTS__/TestConstants/createEntryMockedFS.ts
```diff
@@ -12,6 +12,11 @@ const TEST_FILE_STRUCTURE = {
           'Border.d.ts' : '// Some content'
         }
       },
+      constants: {
```
patrick-tolosa, line 15: Note that this is only needed if you're writing a test (as this is a MOCKED file system).
We can keep it here, but just keep that in mind

src/actions/color/namedcolor.ts
```diff
@@ -0,0 +1,27 @@
+import Action from "../Action";
```
patrick-tolosa, line 1: Should the filename be camelcase?

namedColor?

src/actions/color/rgb.ts
```diff
@@ -0,0 +1,35 @@
+import Action from "../Action";
+import {IAction} from "../../interfaces/IAction";
+import Param from "../../parameters/Param";
+
+/**
+ * @implements IAction
+ */
+class Rgb extends Action implements IAction {
```
patrick-tolosa, line 8: This could in theory be a single class called "Color" right?
And based on the constructor values, create different params?

What do you think?

PR #43, opened 2020-07-09
refactored resize

strausr:


src/actions/resize/Resize.ts
```diff
@@ -1,13 +1,10 @@
+import aspectRatio from './aspectRatio';
 import crop from './crop';
-import scale from './scale';
 import fill from './fill';
+import scale from './scale';
 import minimumPad from './minimumPad';
+import ResizeAction from './ResizeAction';
 
-export {crop, scale, fill, minimumPad};
+export {aspectRatio, crop, fill, scale, minimumPad, ResizeAction};
```
patrick-tolosa, line 8: I'm starting to think we shouldn't export the Actions themselves, as the users will see it on their IDE hints

`Resize.ResizeAction...`

strausr: will remove

PR #44, opened 2020-07-12
Refactored effect to use params

strausr:


src/actions/effect/sepia.ts
```diff
@@ -1,19 +1,15 @@
 import Action from "../Action";
 import {IAction} from "../../interfaces/IAction";
+import {prepareEffect} from "./prepareEffect";
 
 /**
  * @implements IAction
  */
```
patrick-tolosa, line 7: Can we combine all Effect classes to a single class?

it seems like prepareEffect takes care of most of the heavy lifting.

src/actions/effect/prepareEffect.ts
```diff
@@ -0,0 +1,16 @@
+import ParamValue from "../../parameters/ParamValue";
+import Param from "../../parameters/Param";
+
+/**
+ * @param effect
+ * @param level
+ */
+export function prepareEffect(effect: string, level?: number): Param {
```
patrick-tolosa, line 8: Maybe rename to "createEffectParam"?
We used the phrase "prepareParam" to signify "buildParam" for the Border action, so maybe this name will add some confusion later.

PR #46, opened 2020-07-15
Feature/ensure overlay passes compilation tests

patrick-tolosa:
This PR ensures that Overlay passes our compilation tests.


scripts/createEntrypoints.ts
```diff
@@ -11,7 +11,7 @@ const {
 createEntryPointFromESMPath('actions', 'actions');
 
 // Create an Entry Point from everything under bundles/esm/colors -> dist/colors
-createEntryPointFromESMPath('colors', 'colors');
+createEntryPointFromESMPath('constants', 'constants');
```
patrick-tolosa, line 14: 'colors' was a typo, and the actual value should have always been 'constants'

src/actions/layers/Source.ts
```diff
@@ -0,0 +1,15 @@
+import TransformableImage from "../../transformation/TransformableImage";
```
maoznir, line 1: File name should match class name no?

scripts/createEntrypoints.ts
```diff
@@ -11,7 +11,7 @@ const {
 createEntryPointFromESMPath('actions', 'actions');
 
 // Create an Entry Point from everything under bundles/esm/colors -> dist/colors
```
maoznir, line 13: You should also update the comment

__TESTS__/unit/actions/Overlay.test.ts
```diff
@@ -14,12 +14,12 @@ describe('Tests for overlay actions', () => {
       })
       .resize(fill(1000, 1000))
       .overlay(
-        overlayLayer(
-          imageSource('sample')
+        imageLayer(
+          image('sample')
```
maoznir, line 18: Shouldn't we make it possible to import only source.image instead of whole source ?

patrick-tolosa: Yes, we should, I'm not entirely sure what's the bundle impact on keeping it together.

For now, Source seems to be a dumb wrapper around TransforambleImage.

src/actions/layers/Source.ts
patrick-tolosa: Source is not an Entrypoint in itself (Something we might want to discuss)

It's not possible to `import Source from 'base/actions/source'`

instead, it's imported through Layers

It's not possible to `import {Source} from 'base/actions/layers'.

A source on its own is not really an action, so I found it weird having it as an entrypoint.

maoznir: the class ImageSource exists in a file named Source. I would put it in ImageSource.ts file.
unless you want to export ImageSource, TextSource, etc from the Source.ts file

__TESTS__/unit/actions/Overlay.test.ts
maoznir: Lets not have a "for now" for overlay's source at this stage, we should finalize it's spec

maoznir: We will have different types of sources - image,text,video,subtitles etc no? lets allow to import only whats needed

src/actions/layers/Source.ts
patrick-tolosa: For now, I think  that's how it's going to look, I have a feeling that we'll remove Source from the current place any way due to source being used in other, non `overlay` specific contexts

Such as this:

`.effect(Effect.styleTransfer(Source.image('lighthouse')`

I'm honestly not too sure where to place it right now.

__TESTS__/unit/actions/Overlay.test.ts
patrick-tolosa: This is related to the above comment, Obviously that's the goal, but we're missing the right place to put Source right now.

Source is not an action, so I'm wondering where we should place it to make it meaningful export.

src/actions/layers/Source.ts
maoznir: We need a Params folder

maoznir: or under Overlay

PR #48, opened 2020-07-15
updated border action to use enum color or rgb

strausr:


src/actions/border/solid.ts
```diff
@@ -20,14 +20,21 @@ class Solid extends Action implements IBorderAction{
...
+    this.borderColor = this.prepareColor(borderColor);
     return this;
   }
 
+  /**
+   * Returns RGB or Color
+   * @param color
+   */
+  prepareColor(color: string){
+    return color.match(/^#[0-9A-F]{6}$/i) ? `rgb:${color.substr(1)}` : color;
```
patrick-tolosa, line 35: This regex does not support all possible values, right?

`#FFF` - Shorthand for `#FFFFFF`
`#FFFFFF`- 'Reguar' form
`#FFFFFFFF` - The last two F's include the alpha channel for transparency

Are all valid values by us

https://res.cloudinary.com/demo/image/upload/c_pad,w_500,h_100/b_rgb:bbb/sample.png
https://res.cloudinary.com/demo/image/upload/c_pad,w_500,h_100/b_rgb:000fff/sample.png
https://res.cloudinary.com/demo/image/upload/c_pad,w_500,h_100/b_rgb:000fff50/sample.png


In short, shouldn't we just skip this regex and check that the first character is `#`? 
Else we also need to check that the string has 3, 6 or 8 characters (after the `#`)
So either we do a full check, or we just take the first character as a hint

strausr: updated to match with PHP logic and added more tests

PR #50, opened 2020-07-27
Add Gravity param to Base

patrick-tolosa:
Notes for the reviewer:

- I could not find a proper place to put AutoGravity, It has no definite place since it's a very unique Entity. Suggestions are welcomed.
- AutoGravity is not tree-shakable, you'll get all the static functions once you import it. It's very small but still worth noting. The alternative looked very confusing (in terms of imports)
- There are still some issues with the compilation tests, so not all of Gravity could be implemented, however the vast majority is included in this PR.

### DIFF
--- a/__TESTS__/unit/types/Gravity.test.ts
+++ b/__TESTS__/unit/types/Gravity.test.ts
@@ -65,6 +65,6 @@ describe('Gravity Param', () => {
   });
 
   it('Tests Gravity.auto() with AutoGravity with object weighed', () => {
-    expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT, 50)).toString()).toBe('g_auto:cat_50');
+    expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT, -100)).toString()).toBe('g_auto:cat_-100');
   });
 });
--- a/__TESTS__/unit/types/Gravity.test.ts
+++ b/__TESTS__/unit/types/Gravity.test.ts
@@ -1,0 +1,70 @@
+import Gravity from "../../../src/params/gravity/Gravity";
+import * as GravityESM from "../../../src/params/gravity/Gravity";
+
+import * as GravityObjects from "../../../src/constants/gravityObjects/GravityObjects";
+import {AutoGravity} from "../../../src/constants/gravityObjects/GravityObjects";
+
+
+
+describe('Gravity Param', () => {
+  it('Ensure ESM and Default export the same thing', () => {
+    (Object.keys(GravityESM) as Array<keyof typeof GravityESM>) .forEach((funcName) => {
+      if (typeof GravityESM[funcName] === 'function') {
+        // Ensure function exists on both objects
+        expect(GravityESM[funcName]).toEqual( (Gravity as any)[funcName]);
+
+        // sanity, ensure the result is a function at all (and not an accidental undefined)
+        expect(typeof GravityESM[funcName]).toBe('function');
+      }
+    });
+  });
+
+  it('Tests simple gravitation', () => {
+    expect(Gravity.auto().toString()).toBe('g_auto');
+    expect(Gravity.center().toString()).toBe('g_center');
+    expect(Gravity.face().toString()).toBe('g_face');
+    expect(Gravity.faces().toString()).toBe('g_faces');
+    expect(Gravity.east().toString()).toBe('g_east');
+    expect(Gravity.west().toString()).toBe('g_west');
+    expect(Gravity.north().toString()).toBe('g_north');
+    expect(Gravity.south().toString()).toBe('g_south');
+    expect(Gravity.advancedFace().toString()).toBe('g_adv_face');
+    expect(Gravity.advancedFaces().toString()).toBe('g_adv_faces');
+    expect(Gravity.advancedEyes().toString()).toBe('g_adv_eyes');
+    expect(Gravity.northWest().toString()).toBe('g_north_west');
+    expect(Gravity.southEast().toString()).toBe('g_south_east');
+    expect(Gravity.ocrText().toString()).toBe('g_ocr_text');
+  });
+
+  it('Tests gravity objects - Singular', () => {
+    expect(Gravity.object(GravityObjects.CAT).toString()).toBe('g_cat');
+  });
+
+  it('Tests gravity objects - Many', () => {
+    expect(Gravity.object(GravityObjects.CAT, GravityObjects.DOG).toString()).toBe('g_cat:dog');
+  });
+
+  it('Tests Gravity.auto() with objects', () => {
+    expect(Gravity.auto(GravityObjects.CAT, GravityObjects.DOG).toString()).toBe('g_auto:cat:dog');
+  });
+
+  it('Tests Gravity.auto() with AutoGravity.classic', () => {
+    expect(Gravity.auto(AutoGravity.classic()).toString()).toBe('g_auto:classic');
+  });
+
+  it('Tests Gravity.auto() with AutoGravity with object', () => {
+    expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT)).toString()).toBe('g_auto:cat');
+  });
+
+  it('Tests Gravity.auto() with AutoGravity with subject', () => {
+    expect(Gravity.auto(AutoGravity.subject()).toString()).toBe('g_auto:subject');
+  });
+
+  it('Tests Gravity.auto() with AutoGravity with object avoid', () => {
+    expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT).avoid()).toString()).toBe('g_auto:cat_avoid');
+  });
+
+  it('Tests Gravity.auto() with AutoGravity with object weighed', () => {
+    expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT, 50)).toString()).toBe('g_auto:cat_50');
+  });
+});
--- a/scripts/createEntrypoints.ts
+++ b/scripts/createEntrypoints.ts
@@ -13,6 +13,9 @@ createEntryPointFromESMPath('actions', 'actions');
 // Create an Entry Point from everything under bundles/esm/constants -> dist/constants
 createEntryPointFromESMPath('constants', 'constants');
 
+// Create an Entry Point from everything under bundles/esm/params -> dist/params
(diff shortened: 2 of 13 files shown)


scripts/createEntrypoints.ts
```diff
@@ -13,6 +13,9 @@ createEntryPointFromESMPath('actions', 'actions');
 // Create an Entry Point from everything under bundles/esm/constants -> dist/constants
 createEntryPointFromESMPath('constants', 'constants');
 
+// Create an Entry Point from everything under bundles/esm/params -> dist/params
```
maoznir, line 16: Is this related to this pr?

src/constants/gravityObjects/GravityObjects.ts
```diff
@@ -0,0 +1,62 @@
...
+  readonly gravityObject:GravityObject;
+  private weight:number;
+  constructor(gravityObject:GravityObject, weight?: number) {
+    this.weight = weight;
+    this.gravityObject = gravityObject;
+  }
+
+  toString():string {
+    if(this.weight) {
+      const avoidStr = this.weight <=-100 ? 'avoid' : this.weight;
```
maoznir, line 22: 1. what if weight===0?
2. I would put this logic in a separate function and return:
`${name}${getWeight()}

scripts/createEntrypoints.ts
patrick-tolosa: Yes it is, it ensures that the Gravity param is importable by the end user

`import Gravity from '@base/params/gravity'

src/constants/gravityObjects/GravityObjects.ts
patrick-tolosa: 1. What about it really? do we want to skip it? Is cat_0 not a valid input?
2. You'd like to see a function called 'concatNameAndWEight'? 

We could move all this logic (entirely) into a "calculateName or something similar, but then toString() will just call this function., feels a bit off.

wdyt?

maoznir: 1. if(this.weight), will be false for weight===0, that's why i asked, what would you like to happen when weight=0?
2. I expect the void/anything else logic to get more complicated so i'd put the weight logic elsewhere, like calcWeight.

nitzanj: I'm not sure it's well-defined anywhere, but a weight is always a number, `getWeight()` will always return a number. And if the user typed a number, we should keep it as a number in the URL.

The `avoid` keyword is for the other way around - it's there to make the user's life easier. using `avoid` without the need to understand weights. And in that case, we keep the word `avoid` as is. This is a strictly _server-side alias_ that we're not suppose to mess with - we should not replace one for the other on the client-side.

In other words, whatever the user did (either called it with an int weight, or invoked the `avoid()` function), whatever appears in the URL is exactly that - either the weight or the word `avoid`. No need to process anything.
In terms of internal implementation that probably means that you should keep another attribute besides the weight integer, and when building the string look at it first, and if nothing is there use the weight (or nothing at all if weight is null too). 

TL;DR 
`expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT).avoid()).toString()).toBe('g_auto:cat_avoid');`
`expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT, 50)).toString()).toBe('g_auto:cat_50');`
`expect(Gravity.auto(AutoGravity.object(GravityObjects.CAT, -100)).toString()).toBe('g_auto:cat_-100');`

patrick-tolosa: Fixed and added to the tests.

src/constants/gravityObjects/GravityObjects.ts
```diff
@@ -0,0 +1,80 @@
...
+  constructor(name?:string) {
+    this.name = name;
+  }
+
+  toString(): string {
+    return this.name;
+  }
+}
+
+export class AutoGravity {
```
asisayag2, line 12: Why is that located under `constants/...`?

src/actions/resize/aspectRatio.ts
```diff
@@ -1,4 +1,4 @@
-import {ResizeAction} from "./Resize";
+import ResizeAction from "./ResizeAction";
```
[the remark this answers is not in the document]
patrick-tolosa: Spotted it while working, it caused an infinite loop on build time, so I fixed it in this PR,
Can be removed if critical (technically it's a bug fix)

src/constants/gravityObjects/GravityObjects.ts
```diff
@@ -0,0 +1,80 @@
...
+  private shouldAvoid:boolean;
+
+  constructor(gravityObject:GravityObject, weight?: number | string) {
+    this.weight = weight;
+    this.gravityObject = gravityObject;
+    this.shouldAvoid = false;
+  }
+
+  private shouldAddWeight() {
+    return typeof this.weight === 'number' || typeof this.weight === 'string' || this.shouldAvoid;
```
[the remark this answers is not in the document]
patrick-tolosa: There are several perspectives here.

From our perspective, weight is always a number.
From our customer's perspective, weight can be a number or a string (think of databases saving things as strings)

On top of that, a customer can call `.avoid()`, which automatically adds the word `avoid` as weight. `auto:cat_avoid`


Also, accepting a string is a type of future proof for `auto:cat_newOption`

src/constants/gravityObjects/GravityObjects.ts
```diff
@@ -0,0 +1,80 @@
...
+    }
+  }
+
+  toString():string {
+    // Future proofing, in case we'd like to support some custom string in the future, or if data is coming from a DB
+    if(this.shouldAddWeight()) {
+      return `${this.getName()}_${this.getWeight()}`;
+    } else {
+      return this.getName();
+    }
```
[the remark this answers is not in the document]
patrick-tolosa: I guess, I personally prefer this syntax for readability, but open for change.

src/constants/gravityObjects/GravityObjects.ts
```diff
@@ -0,0 +1,80 @@
...
+
+export const BIRD = new GravityObject('bird');
+export const DOG = new GravityObject('dog');
+export const CAT = new GravityObject('cat');
+export const MICROWAVE = new GravityObject('microwave');
+export const REFRIGERATOR = new GravityObject('refrigerator');
+export const BOTTLE = new GravityObject('bottle');
+export const SINK = new GravityObject('sink');
+export const SKATEBOARD = new GravityObject('skateboard');
+export const PERSON = new GravityObject('person');
```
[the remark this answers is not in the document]
patrick-tolosa: The Enums for GravityObjects, but I'm sure that's not what you meant.

__TESTS__/unit/types/Gravity.test.ts
```diff
@@ -0,0 +1,70 @@
+import Gravity from "../../../src/params/gravity/Gravity";
+import * as GravityESM from "../../../src/params/gravity/Gravity";
+
+import * as GravityObjects from "../../../src/constants/gravityObjects/GravityObjects";
+import {AutoGravity} from "../../../src/constants/gravityObjects/GravityObjects";
+
+
+
+describe('Gravity Param', () => {
+  it('Ensure ESM and Default export the same thing', () => {
```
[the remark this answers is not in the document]
patrick-tolosa: Sure,

GravityESM contains all the possible imports from Gravity in a single Object.
Our users will usually want to import only what they need 

`import {object} from 'base/params/gravity'`.

However, if you want to import everything inside `params/gravity` and not list every single function, you can import with a `*`

What we test here is that everything that's exported in the default way can also be imported as an individual module.


It's possible that `import Gravity from '...';  Gravity.object` to exist, but 
`import {object} from '...'` to fail. 

Default import does not have to contain the same objects as Named imports
This test ensures that we didn't forget anything (And won't forget anything in the future).


As a reference, we've had issues with this in Cloudinary-core, where a named export would exist, but a default export would not (or vice verca)

src/constants/gravityObjects/GravityObjects.ts
patrick-tolosa: It's a very good question (and part of my notes to the reviewer).

Imports in JS are not like imports in Java or PHP, we have to explicitly import it from somewhere, we don't have namespaces.


There's simply no logical place to put AutoGravity, as there are no similar Entities.
For example, if this was part of Gravity, the code would look like this:

`Gravity.auto(Gravity.AutoGravity.object(GravityObjects.CAT))` - Not ideal.

maoznir: I suggest that things that are related to gravity will be in a Gravity folder, in folders like params, gravityObjects etc

patrick-tolosa: That would indeed be great, but how would a user import it? 
Unlike JAVA we can't just import the 'Gravity' namespace.

We also need to support granular imports, a user should only be able to pull only the constants he wants.
It's a lot of things that really limit how flexible we can be with the file structure.

How do you see the whole solution?

src/actions/resize/aspectRatio.ts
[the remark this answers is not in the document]
asisayag2: I always prefer separating PRs

src/constants/gravityObjects/GravityObjects.ts
asisayag2: Sure that there are not other places that such animals exists? Something that will justify creating a separated folder  structure for? 
I would never look for it under `constants`

src/constants/gravityObjects/GravityObjects.ts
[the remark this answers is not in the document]
asisayag2: So... why not 
`return this.weight!==null || this.shouldAvoid`
?

Another option is delegating the entire weight handling to the `getWeight` method.
In the `toString` just return `${this.getName()}${this.getWeight()}`
and in the `getWeight` return `'_avoid' / '' / '_${this.weight}` according to what is sent

src/constants/gravityObjects/GravityObjects.ts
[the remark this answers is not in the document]
asisayag2: It's supposed to be a huge enum. maybe have it in its own file?

PR #51, opened 2020-07-28
Validate the contents of /dist when we create entrypoints

patrick-tolosa:
This PR deals with validating the contents of the `/dist` folder.

At the moment, errors in output caused by faulty scripts or human errors are possible.
This PR adds automation to this process but ensuring that dist contains exactly the expected files.

First change - return created file structure from `createEntryPointFromESMPath`

Second change - Create a function that given an input file structure, validates that the disk file structure contains the exact match.

We check for two things:
1. All files in the input object exist in the file system
2. All files in the file system exist in the input object

These checks ensure that no unexpected files are leaking to the /dist folder, and that the contents are as expected.
The two errors above can be suppressed using a configuration, these options can be see in the tests


__TESTS__/unit/Instance/Cloudinary.test.ts
```diff
@@ -17,6 +17,25 @@ describe('Tests for Cloudinary instance', () => {
 
     expect(tImage.toURL()).toBe('http://res.cloudinary.com/demoInInstance/image/upload/c_fill,h_10,w_10/sample');
   });
+
+
+  it ('Can pass config in runtime instead of in creation', () => {
```
patrick-tolosa, line 22: This file was changed to increase code coverage.

scripts/lib/entryPointsLib.ts
```diff
@@ -51,7 +55,7 @@ function createEntryPointFromESMPath(pathInESMBundle: string, entryPointInDist:
     const isDir = fs.lstatSync(`${FULL_SOURCE_PATH}/${name}`).isDirectory();
 
     if (isDir) {
-      fs.mkdirSync(`${FULL_ENTRY_POINT_PATH}/${name}`);
+      fs.mkdirSync(`${FULL_ENTRY_POINT_PATH}/${name}`, {recursive: true});
```
patrick-tolosa, line 58: This change prevents node from crashing if the directory already exists

scripts/validatePackageFiles.ts
```diff
@@ -0,0 +1,116 @@
+/* eslint-disable @typescript-eslint/no-var-requires */
+const fs = require('fs');
+const path = require('path');
+const MissingRequiredFile = require('./errors/MissingRequiredFile');
+const FoundNonRequiredFile = require('./errors/FoundNonRequiredFile');
+
+interface IFileStructureLevel{
```
maoznir, line 7: Maybe i don't quite understand bu:
1. I'd move this to a file of its own, like FileStructure, FileStructureLevel
2. How does a file structure tree look like?
3. Isn't there a representation of folders and files tree, that we can use instead of 'reinventing the wheel'?

scripts/createEntrypoints.ts
```diff
@@ -1,23 +1,47 @@
 /* eslint @typescript-eslint/no-var-requires: 0 */
```
maoznir, line 1: Please add documentation of what this file contains, what its used for

scripts/validatePackageFiles.ts
maoznir: also, is this script used only for tests? if so, please move to the tests folder

scripts/createEntrypoints.ts
patrick-tolosa: Added a description

scripts/validatePackageFiles.ts
patrick-tolosa: A few things: 

1. We can move interfaces to their own files, it's just the interfaces though, so that felt a bit strange
2. The shape is included in the interface, it's a key/value pair of anything you want, specifically, we allow a flag and the end value can be either a string or a boolean.
3. This is (almost) the same file structure representation that's used by Mock-FS (Which we use)

This script is not used only for tests, it's used to validate the output of the `createEntryPoints` script

PR #55, opened 2020-08-02
Implement Position param

patrick-tolosa:
Implemented Position as a Param for Overlay.

### DIFF
--- a/src/actions/layers/Layers.ts
+++ b/src/actions/layers/Layers.ts
@@ -2,29 +2,65 @@ import {ILayerAction} from "./ILayerAction";
 import Action from "../Action";
 // TODO - BundleSize Warning - we include all the Sources code within Layers.
 import Source, {ImageSource} from "./Source";
+import {Position} from "../../params/position/Position";
+import Param from "../../parameters/Param";
 
 
 class Layer extends Action implements ILayerAction {
   source: ImageSource; // TODO this needs to accept other types of sources
-  constructor(transformable: ImageSource) {
+  position:Position;
+  constructor(transformable: ImageSource, position:Position) {
     super();
     this.source = transformable;
+    this.position = position;
+  }
+
+  /**
+   * Layers are built using three bits -> /Open/Transform/Close
+   * The opening of a layer
+   */
+  openLayer() {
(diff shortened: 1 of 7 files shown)


src/actions/layers/Layers.ts
```diff
@@ -2,29 +2,65 @@ import {ILayerAction} from "./ILayerAction";
...
     this.source = transformable;
+    this.position = position;
+  }
+
+  /**
+   * Layers are built using three bits -> /Open/Transform/Close
+   * The opening of a layer
+   */
+  openLayer() {
+    return `l_${this.source.asset.publicID}`;
```
nitzanj, line 23: This only supports regular image layers - when adding video/fetch/whatever this function will look differently I assume.

src/actions/layers/Layers.ts
```diff
@@ -2,29 +2,65 @@ import {ILayerAction} from "./ILayerAction";
...
+  layerTransformation() {
+    return this.source.toString();
+  }
+
+  /**
+   * Layers are built using three bits -> /Open/Transform/Close
+   * Closing of the layer, includes Position as well
+   */
+  closeLayer() {
+    const bit = new Action().addParam(new Param('fl', 'layer_apply'));
```
nitzanj, line 39: Some actions use layer and utilize falgs (Cutter action is using fl_cutter) - are the flags handled in the `addParam` function? otherwise you need to make sure the flags are concatenated (`fl_cutter.layer_apply`).

src/actions/layers/Layers.ts
patrick-tolosa: Yes this is specific to images, videos will be added and changes will be done where needed.

src/actions/layers/Layers.ts
patrick-tolosa: If these flags are all added on the layer action, this shouldn't be a problem.

nitzanj: The flags are part of the layer action itself, it should be self-contained.
If the user wants to add a flag to the resulting action at some later point that's a a different story and unrelated anyway (and also pretty much undefined when it comes to layers...).

LGTM

PR #57, opened 2020-08-04
Add all missing Resize types

patrick-tolosa:
- Add all missing types to Resize
- Add height() and width()
- Set constructor params to optional
- Makes Resize tests more reusable


src/actions/resize/IResizeAction.ts
```diff
@@ -4,4 +4,6 @@ import {GravityParam} from "../../params/gravity/Gravity";
 export interface IResizeAction extends IAction{
   aspectRatio(ratio:string | number):IResizeAction;
   gravity(type:GravityParam):IResizeAction;
+  height(x:number):IResizeAction;
```
strausr, line 7: Cant also be a string?

strausr, line 7: Can't this also be a string?

patrick-tolosa: Yea I think it can actually, but height and width. and the constructor values as well

PR #59, opened 2020-08-05
Implement blendmode for overlays

patrick-tolosa:
- Added a Utility to match ESM to default packages (expectESMToMatchDefault)
- Added implementation needed for BlendMode
- Refactored some tests to reduce bloat in the amount of lines

### DIFF
--- a/__TESTS__/TestUtils/expectESMToMatchDefault.ts
+++ b/__TESTS__/TestUtils/expectESMToMatchDefault.ts
@@ -1,0 +1,24 @@
+/**
+ * @description given an ESM package and a default export, this function expects the two to be identical
+ * Example usage:
+ *    import * as GravityESM from '/params/gravity/Gravity'
+ *    import Gravity from '/params/gravity/Gravity'
+ *
+ *    it ('Should work', () => {
+ *       expectESMToMatchDefault(GravityESM, Gravity);
+ *    })
+ *
+ * @param {object} ESM
+ * @param {object} Default
+ */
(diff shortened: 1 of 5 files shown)


src/params/blendMode/BlendMode.ts
```diff
@@ -0,0 +1,50 @@
...
+
+/**
+ * @description Add an overlay image blended using the 'overlay' blend mode.
+ */
+function overlay():BlendMode {
+  return new BlendMode('overlay');
+}
+
+export {
+  screen,
```
maoznir, line 39: I think blend mode should include: multiply screen mask anti_removal

__TESTS__/TestUtils/expectESMToMatchDefault.ts
```diff
@@ -0,0 +1,24 @@
+/**
+ * @description given an ESM package and a default export, this function expects the two to be identical
```
maoznir, line 2: should be part of a different pr?

src/params/blendMode/BlendMode.ts
patrick-tolosa: This is aligned with Kotlin, 

Mask is not documented and believed to be an error in the spec
anti_removal is snoozed for later discussion

__TESTS__/TestUtils/expectESMToMatchDefault.ts
patrick-tolosa: I had to write all this code for the tests, all I did was just put it in a different function. 
I believe it belongs here

PR #64, opened 2020-08-12
Align round-corners-to-spec

patrick-tolosa:
#### What does this PR solve?
This PR aligns RoundCorners `r_` with our specifications.
In this PR we didn't use an interface `IRoundCorners`, and in general, they seem to be redundant, we might consider removing them in the future

#### Final checklist


src/actions/roundCorners/RoundCornersAction.ts
```diff
@@ -0,0 +1,43 @@
...
+  }
+
+  /**
+   * @param {number} a
+   * @param {number} b
+   * @param {number} c
+   * @param {number} d
+   * @return {RoundCornersAction}
+   */
+  radius(a:number, b?:number, c?:number, d?:number):this {
```
maoznir, line 23: Why not meaningful names?

patrick-tolosa: It's four identical radiuses, so..

radiusA, radiusB, radiusC, radiusD?

maoznir: As a user, how do I know how each one will affects the image? shouldn't their names and docstring convey that?

patrick-tolosa: Added the PHP Docstrings and links to the reference

PR #66, opened 2020-08-16
Added conditionals

strausr:
#### What does this PR solve?
Added conditional action

#### Final checklist
- [*] JSdoc - Add proper docstrings based on our PHP implementation
- [*] JSdoc - Hide private functions using @private
- [*] Implementation is aligned with PHP Reference(If different, please specify why)
- [*] Tests - Add proper tests to the added code


src/actions/condition/Condition.ts
```diff
@@ -0,0 +1,49 @@
...
+      .map((val) => CONDITIONAL_OPERATORS[val] || val)
+      .join('_');
+
+
+    return resultExpression;
+  }
+}
+
+/**
+ * @param key
```
patrick-tolosa, line 41: Please add the missing JSDoc (if we already know what they are from PHP 2.0)

If we don't know what they should be, please open an internal issue for future work to ensure the docstrings are added.

strausr: updated with what is know and opened a ticket for unknowns

__TESTS__/unit/actions/Condition.test.ts
```diff
@@ -0,0 +1,117 @@
...
+  });
+
+  it('Creates a cloudinaryURL with condition =', () => {
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .ifCondition(Condition.fromString('ar = 1.0'))
+      .setPublicID('sample')
+      .toURL();
+
+    expect(url).toBe('http://res.cloudinary.com/demo/image/upload/if_ar_eq_1.0/sample');
```
patrick-tolosa, line 28: This transformation isn't valid, can we make sure our tests have valid and correct usages?
(this, for example, is missing an end-if)

Tests become a part of our informal documentation as to what's allowed and what's not

strausr: updated

PR #70, opened 2020-08-19
added effect colorize, oilpaint, shadow

strausr:
#### What does this PR solve?
added effect colorize, oilpaint, shadow

#### Final checklist

### DIFF
--- a/src/actions/effect/oilPaint.ts
+++ b/src/actions/effect/oilPaint.ts
@@ -3,7 +3,7 @@ import {IEffectAction} from "./IEffectAction";
 
 
 /**
- * @description Applies a oilPaint filter to the asset.
+ * @description Applies an oilPaint filter to the asset.
  * @memberOf Actions.Effect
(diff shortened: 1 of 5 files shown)


src/actions/effect/oilPaint.ts
```diff
@@ -0,0 +1,14 @@
+import EffectAction from "./EffectAction";
+import {IEffectAction} from "./IEffectAction";
+
+
+/**
+ * @description Applies a oilPaint filter to the asset.
```
patrick-tolosa, line 6: Shouldn't it be "an" oilpaint?

patrick-tolosa (approved):

patrick-tolosa:
Please add the right types in the JSDoc strings, but other than that looks good

src/actions/effect/oilPaint.ts
strausr: yas

PR #75, opened 2020-08-23
Refactor resize to accept string as an argument

patrick-tolosa:
#### What does this PR solve?
This PR allows all the resize functions to accept strings as well as numbers.
An attempt was made to reuse all the type definitions instead of adding changing them one by one.

#### Commit Changes:
* Create a type for all resize actions (resizeActionFactory type)
* Consolidate all type definitions to the factory type
* Remove all type defintions from the resize functions
* Change all docstrings to be number|string
* Add a Types namespace to put all our types
* Remove of the IResizeAction (It seemed redundant), if we'll need it in the future we can re-add it but for now it looks like it's pointless

We still have an open question on how to reuse all the "param" definitions

#### Final checklist


strausr (approved):

__TESTS__/unit/actions/Resize.test.ts
```diff
@@ -4,7 +4,7 @@ import CloudinaryConfig from "../../../src/config/CloudinaryConfig";
 import {AutoGravity} from "../../../src/constants/gravityObjects/GravityObjects";
 import Gravity from "../../../src/params/gravity/Gravity";
 import * as GravityObjects from '../../../src/constants/gravityObjects/GravityObjects';
-import {IResizeAction} from "../../../src/actions/resize/IResizeAction";
+import {IResizeAction} from "../../../src/actions/resize/IResizeTypes";
```
maoznir, line 7: the descriptions says this was removed?

src/actions/resize/IResizeTypes.ts
```diff
@@ -0,0 +1,16 @@
+import ResizeAction from "./ResizeAction";
+
+/**
+ * @memberOf Types
+ * @name resizeActionFactory
+ * @description
+ * This is the type for all resize action methods (scale, fit, crop, etc.).
+ */
+export type resizeActionFactory = (width?:number|string, height?:number|string) => ResizeAction
```
maoznir, line 9: Why the word "factory" do we use this for other builders?

patrick-tolosa: This is a good question.
Strictly speaking, these functions are factories that return a Builder Class for an action.

Resize has a lot of these functions, and they're all reused, so it made sense to extract all these declarations into one place.

We could create these types for all of our action functions, but if they don't share the signature then it makes no sense.

The idea was the share the signature of the resizeAction(factory)
What do you think?

__TESTS__/unit/actions/Resize.test.ts
patrick-tolosa: The interface for ResizeAction was removed (initially we had an interface for every action, but it wasn't really needed).

We do have some Types left though that are needed. (If we decide to keep the resizeActionFactory)

patrick-tolosa: Good catch, removed.

src/actions/resize/IResizeTypes.ts
maoznir: As long as it's internal only and does not affect the user im ok with it

maoznir (approved):

PR #77, opened 2020-08-24
Add action background

strausr:
#### What does this PR solve?
Adds action background

#### Final checklist


src/actions/background/Background.ts
```diff
@@ -0,0 +1,18 @@
+/**
+ * Defines the background color to use instead of transparent background areas or when resizing with padding.
+ *
+ * **Learn more**:
+ * <a href="https://cloudinary.com/documentation/image_transformations#setting_background_color" target="_blank">
+ * Setting background for images</a> |
+ * <a href="https://cloudinary.com/documentation/video_manipulation_and_delivery#background_color" target="_blank">
+ * Setting background for videos</a>
+ *
+ * @api
```
patrick-tolosa, line 10: Unfortunately, we can't just copy the docstrings from PHP, certain things need to change.

Also, please make sure this looks proper when you build the docs (npm run build:docs).
Pay attention to line breaks and spaces.

Also, `Background` should be part of the Actions namespace, when building the docs ensure there's an `Actions.Background` entry (under namespaces)

src/actions/background/IBackgroundAction.ts
```diff
@@ -0,0 +1,5 @@
+import {IAction} from "../../interfaces/IAction";
+
+export interface IBackgroundAction extends IAction{
```
patrick-tolosa, line 3: I know we used these interfaces before, but I'd like to question if we really need them.
Can this be removed?

src/actions/background/auto.ts
```diff
@@ -0,0 +1,11 @@
+import BackgroundAction from "./BackgroundAction";
+import {IBackgroundAction} from "./IBackgroundAction";
+
+/**
+ * Applies background color automatically.
+ */
+function auto(): IBackgroundAction{
```
patrick-tolosa, line 7: Functions without a namespace applied to them (in the docstring) will appear as public in the JSdoc,
This may or may not be what you wanted to implement.

src/actions/resize/IResizeAction.ts
```diff
@@ -1,9 +1,11 @@
 import {IAction} from "../../interfaces/IAction";
 import {GravityParam} from "../../params/gravity/Gravity";
+import {IBackgroundAction} from "../background/IBackgroundAction";
 
 export interface IResizeAction extends IAction{
```
patrick-tolosa, line 5: The IResizeAction is scheduled for removal in a different PR, can you please undo the changes here?

patrick-tolosa (changes requested):
Needs adjustments for JSdoc strings to ensure things are placed properly.

src/actions/background/BackgroundAction.ts
```diff
@@ -0,0 +1,20 @@
...
+import Param from "../../parameters/Param";
+import ParamValue from "../../parameters/ParamValue";
+
+class BackgroundAction extends Param implements IAction {
+  constructor(backgroundType?: string) {
+    super('b', new ParamValue(backgroundType).setDelimiter('_'));
+  }
+
+  /**
+   * @memberOf Actions.Background
```
patrick-tolosa, line 11: This method is not actually a member of Actions.Background right?

A user cannot import {contract} from actions/background.

patrick-tolosa (approved):

src/actions/background/BackgroundAction.ts
strausr: But it is public. How should we add it to the docs?

PR #84, opened 2020-08-26
Add action flag

strausr:
#### What does this PR solve?
Add action flag

#### Final checklist


__TESTS__/unit/actions/Flag.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  }
+});
+
+describe('Tests for Transformation Action -- Flag', () => {
+  it('Expects ESM to match Default', () => {
+    expectESMToMatchDefault(FlagESM, Flag);
+  });
+
+  it('Ensure namespace is correctly populated', () => {
+    expect(Flag.animated).toEqual(animated);
```
patrick-tolosa, line 20: This line isn't needed anymore, it's covered by line 16

patrick-tolosa (approved):

nitzanj (commented):
This looks good, two questions " 
* How we'll be able to reuse - these are all `FlagAction` and some flags may be used as part of other actions. I see fl_layer_apply and fl_cutter are not here, presumably because they're used differently as part of other actions - but that means that whoever is using them is not reusing the flag code here (so a bit of string templating is duplicated - `f_whatever`).

* Who's in charge of the dot concatenation? e.g. `fl.layer_appy.cutter`? does this happen when generating the string? We need to make sure that flags don't overwrite each other, as they're the only param that can exist in the same component more than once.

patrick-tolosa:
> This looks good, two questions "
> 
> * How we'll be able to reuse - these are all `FlagAction` and some flags may be used as part of other actions. I see fl_layer_apply and fl_cutter are not here, presumably because they're used differently as part of other actions - but that means that whoever is using them is not reusing the flag code here (so a bit of string templating is duplicated - `f_whatever`).
> * Who's in charge of the dot concatenation? e.g. `fl.layer_appy.cutter`? does this happen when generating the string? We need to make sure that flags don't overwrite each other, as they're the only param that can exist in the same component more than once.

@nitzanj  Good point there,

This is the way we use fl_layer_apply
```
...new Param('fl', 'layer_apply')
```
Clearly this will come up with your Cutter example as something that will require change. 
I'm not sure it's blocking for now, as it'll be addressed later.

Regarding your second question, that entire syntax is currently residing on the `Action` itself, through a method called `addFlag`

```javascript
  addFlag(flag: Flag): this {
    const existingFlag = this.params.get('fl_');
    flag.paramValue.setDelimiter('.');

    if (existingFlag){
      existingFlag.addValue(flag.paramValue);
    } else {
      this.params.set('fl_', flag);
    }

    return this;
  }
```

If we're moving this functionality to the Transformation Level, this needs to be properly addressed.
@strausr we should have a test that tests multiple flags on the ImageLevel

I'm not entirely sure what you mean about the same flag existing in the same component, surely something like this can't exist, right?
`/fl_foo.foo.foo/` or `fl_foo,fl_foo`

__TESTS__/unit/actions/Flag.test.ts
```diff
@@ -0,0 +1,232 @@
...
+  });
+
+  it('Creates a cloudinaryURL with Flag animated', () => {
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .addFlag(Flag.animated())
+      .setPublicID('sample')
```
patrick-tolosa, line 27: We should add a test for multiple flags, these should be concatenated with a dot.

such as `fl_foo.bar.zoo`

src/params/flag/Flag.ts
```diff
@@ -0,0 +1,53 @@
...
+import immutableCache from "./immutableCache";
+import keepAttribution from "./keepAttribution";
+import keepIptc from "./keepIptc";
+
+/**
+ * @description Defines flags that you can use to alter the default transformation behavior.
+ * @namespace Flag
+ * @memberOf Params
+ */
+class Flag extends Param {
```
patrick-tolosa, line 32: I would move this to a separate file as this creates a circular dependency import

Flag imports from getInfo, which imports from Flag. 
Lets just move this decleration somewhere else.

patrick-tolosa (approved):

PR #86, opened 2020-08-26
Add underlay support

strausr:
#### What does this PR solve?
Added support for underlay in Layer
Removed use of ILayer

### DIFF
--- a/src/actions/layers/Layers.ts
+++ b/src/actions/layers/Layers.ts
@@ -5,7 +5,6 @@
  */
 
 
-import {ILayerAction} from "./ILayerAction";
 import Action from "../Action";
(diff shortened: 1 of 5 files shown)


src/actions/layers/Layers.ts
```diff
@@ -15,23 +14,32 @@ import {BlendMode} from "../../params/blendMode/BlendMode";
...
     this.source = transformable;
     this.position = position;
     this.blendMode = blendMode;
   }
 
+  /**
+   * Sets the layerType with u | l depending if underlay or overlay
+   * @param type
+   */
+  setLayerType(type: string){
```
patrick-tolosa, line 33: can we add stronger types?

`type : 'u' | 'l'`?

patrick-tolosa (approved):

PR #88, opened 2020-08-27
Add Effect Style Transfer

patrick-tolosa:
#### What does this PR solve?
- Create StyleTransfer action
- Extend Layer to support StyleTransfer
- Fix a few type errors in Layer

#### Final checklist
  Implementation is more 'JS friendly', however compilation tests work 'as-is'

### DIFF
--- a/__TESTS__/unit/actions/Effect.test.ts
+++ b/__TESTS__/unit/actions/Effect.test.ts
@@ -6,6 +6,7 @@ import * as EffectESM from "../../../src/actions/effect/Effect";
 import Effect from "../../../src/actions/effect/Effect";
 import * as Outline from "../../../src/constants/outline/Outline";
 import {image} from "../../../src/params/sources/Sources";
+import scale from "../../../src/actions/resize/scale";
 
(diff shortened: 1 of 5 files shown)


__TESTS__/unit/actions/Effect.test.ts
```diff
@@ -260,4 +261,37 @@ describe('Tests for Transformation Action -- Effect', () => {
 
     expect(url).toBe('http://res.cloudinary.com/demo/image/upload/e_outline:outer:20:200/sample');
   });
+
+  it('Creates a cloudinaryURL with effect style_transfer', () => {
+    const url = new TransformableImage()
+      .setConfig(CONFIG_INSTANCE)
+      .effect(Effect.styleTransfer(image('woman')))
```
nitzanj, line 268: Can a transformation be applied to the source for this effect? Looking at the code for `ImageSource` I think it should work, but we need to verify and add a test.

patrick-tolosa: Added as a test, the end result is, for example, 

```
l_woman/c_scale,h_100,w_100/e_style_transfer:preserve_color:15,fl_layer_apply/sample
```

PR #89, opened 2020-08-27
Fix ar_1.0 turning into ar_1

patrick-tolosa:
#### What does this PR solve?
Fixes a type issue with whole numebrs passed to aspectRatio

When 1.0 is passed, JS turns it into 1.
so aspectRatio(1.0) turns into ar_1, which is incorrect.

This PR always adds a decimal point

#### Final checklist
- [ ] JSdoc - Add proper docstrings based on our PHP implementation

### DIFF
--- a/__TESTS__/unit/actions/Resize.test.ts
+++ b/__TESTS__/unit/actions/Resize.test.ts
@@ -188,4 +188,16 @@ describe('Tests for Transformation Action -- Resize', () => {
     const url = getImageWithResize(Resize.scale(100).x(1).y(2).zoom(3), 'url');
     expect(url).toContain('c_scale,w_100,x_1,y_2,z_3');
   });
+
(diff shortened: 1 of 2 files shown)


patrick-tolosa:
@nitzanj  - Should AR always be a decimal?

nitzanj:
@patrick-tolosa ar_2 and ar_2.0 is identical in terms of actual result, however for consistency we need to make sure all SDKs output the same url-strings for identical cases. 
I'm pretty sure we've already decided on the correct way, and I'm pretty sure it was indeed to always treat AR as float (so ar_1.0). But I can't find the reference. @const-cloudinary am I remembering right? I can't find it written anywhere.

patrick-tolosa:
@nitzanj  this was (prematurely) merged, I was sure this PR addresses that consistency.
I'll make sure this gets addressed.

PR #98, opened 2020-09-03
Align Adjust to spec

patrick-tolosa:
#### What does this PR solve?
Align Adjust to Spec

- Remove tint (Not in Spec)
- Add ImproveMode Param (Enum of two values)
- Add recolor()
- Add replaceColor
- Add fillLight()
- Fix improve()
- Add a fix to ParamValue to detect falsish values  correctly (such as 0)

#### Final checklist
- [X] Implementation is aligned with Spec


src/actions/adjust/Adjust.ts
```diff
@@ -215,10 +208,37 @@ function vibrance(strength?:number): EffectAction {
...
+function replaceColor(toColor:string): ReplaceColorAction {
+  return new ReplaceColorAction(toColor);
+}
+
+/**
+ * @description Converts the colors of every pixel in an image based on the supplied color matrix, in which the value of each color channel is calculated based on the values from all other channels (e.g. a 3x3 matrix for RGB, a 4x4 matrix for RGBA or CMYK, etc).</br>
+ * For every pixel in the image, take each color channel and adjust its value by the specified values of the matrix to get a new value.
+ * @param {number[][]} matrix
+ */
+function recolor(matrix:number[][]): RecolorAction {
```
strausr, line 226: I know this is not in the spec but does it not look weird that recolor is not in camalcase?

src/params/improveMode/ImproveMode.ts
```diff
@@ -0,0 +1,2 @@
+export const OUTDOOR = 'outdoor';
```
strausr, line 1: Should this be under constants rather than params?

strausr (changes requested):

src/params/improveMode/ImproveMode.ts
patrick-tolosa: I'm not sure it makes sense to keep both Constants and Params, one of them has to go (And we also need a better name).

I think everything that comes from the "outside" needs to go Params

src/actions/adjust/Adjust.ts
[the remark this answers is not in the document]
patrick-tolosa: As per spec, let's open a bug(or change request) to the spec if you feel it's needed.

__TESTS_BUNDLE_SIZE__/bundleSizeTestCases.ts
```diff
@@ -22,8 +22,8 @@ const bundleSizeTestCases:ITestCase[] = [
     ]
   },
   {
-    name: 'Tests a transformable image with Resize and just',
-    sizeLimitInKB: 11,
+    name: 'Tests a transformable image with Resize and ajust',
```
maoznir, line 25: change to "Adjust"?

patrick-tolosa: Done, also capitzlized Border and Quality.

strausr (approved):

maoznir (approved):

PR #99, opened 2020-09-03
updated conditionals jsdocs

strausr:
#### What does this PR solve?
updated conditionals jsdocs

#### Final checklist

### DIFF
--- a/src/actions/condition/Condition.ts
+++ b/src/actions/condition/Condition.ts
@@ -18,6 +18,11 @@ const CONDITIONAL_OPERATORS: Record<string, string> = {
   "^": "pow"
 };
 
+/**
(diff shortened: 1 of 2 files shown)


src/actions/condition/Condition.ts
```diff
@@ -18,6 +18,11 @@ const CONDITIONAL_OPERATORS: Record<string, string> = {
   "^": "pow"
 };
 
+/**
+ * Sets up a conditional transformation.
+ * @memberOf Actions
```
patrick-tolosa, line 23: Is this Class a member of Actions?

Usually, the members of actions are the functions we use, such `scale()` or `fit()`

patrick-tolosa (approved):

PR #103, opened 2020-09-03
Align border to spec

patrick-tolosa:
#### What does this PR solve?
- Add type() method instead of solid
- Pass borderStyle dynamically (instead of hardcoded to slid)

#### Final checklist


__TESTS__/unit/actions/Border.test.ts
```diff
@@ -5,7 +5,7 @@ import CloudinaryConfig from "../../../src/config/CloudinaryConfig";
 import * as Colors from "../../../src/constants/colors/Colors";
 import expectESMToMatchDefault from "../../TestUtils/expectESMToMatchDefault";
 
-const {solid} = Border;
+const {type, SOLID} = Border;
```
maoznir, line 8: this should be borderType no? otherwise it might collide with other names

maoznir (changes requested):

__TESTS__/unit/actions/Border.test.ts
patrick-tolosa: All these changes are aligned to the Spec,
We can open a bug about it in the spec, but this is the agreement we have at the moment (align to the spec, discuss issues later)

strausr (approved):

maoznir (approved):

PR #105, opened 2020-09-03
Align addVariableAction

patrick-tolosa:
#### What does this PR solve?
- Rename variable to addVariable
- Rename Variable.variable to Variable.set
- remove useless interface
- Separated VariableAtion to separate file.

#### Final checklist


src/actions/variable/Variable.ts
```diff
@@ -4,34 +4,18 @@
...
-}
 
 /**
  * @description Defines an new user variable.
  * @memberOf Actions.Variable
  * @param {string} key
  * @param {number | string} value
  */
-function variable(key: string, value: number | string): VariableAction {
+function set(key: string, value: number | string): VariableAction {
```
maoznir, line 16: I think you should change this to a meaningful name like setVariable

maoznir (changes requested):

src/actions/variable/Variable.ts
patrick-tolosa: All these changes are aligned to the Spec,
We can open a bug about it in the spec, but this is the agreement we have at the moment (align to the spec, discuss issues later)

strausr (approved):

maoznir (approved):

PR #122, opened 2020-09-15
Feature/align dither and shadow

patrick-tolosa:
#### What does this PR solve?
This PR aligns Dither and Shadow from Effect.

#### Final checklist

### DIFF
--- a/src/constants/dither/Dither.ts
+++ b/src/constants/dither/Dither.ts
@@ -1,0 +1,20 @@
+export const THRESHOLD_1X1 = 0;
+export const CHECKERBOARD_2X1 = 1;
+export const ORDERED_2X2 = 2;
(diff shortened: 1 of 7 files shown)


src/constants/dither/Dither.ts
```diff
@@ -0,0 +1,20 @@
+export const THRESHOLD_1X1 = 0;
```
strausr, line 1: need to add to index

strausr, line 1: Need to add Dither.ts to index

strausr (changes requested):

src/constants/dither/Dither.ts
patrick-tolosa: As discussed, for now we won't export it in index.ts

PR #145, opened 2020-09-22
Realign blackwhite and add deshake

patrick-tolosa:
#### What does this PR solve?
Realign blackWhite and add Deshake

#### Final checklist


src/actions/effect/Effect.ts
```diff
@@ -18,7 +18,7 @@ import outline from "./outline";
 import styleTransfer from "./styleTransfer";
 import boomerang from "./simple/boomerang";
 import advancedRedEye from "./simple/advanedRedEye";
-import blackWhite from "./simple/blackWhite";
+import blackWhite from "./leveled/blackWhite";
```
strausr, line 21: maybe just level/blackwhite

leveled has a connotation of being even

strausr (approved):

src/actions/effect/Effect.ts
patrick-tolosa: maybe as part of a different PR, we need a good name through

PR #165, opened 2020-10-01
Align docs for all effects

patrick-tolosa:
#### What does this PR solve?
Add JSdoc to all effect actions

#### Final checklist
- [ ] Tests - Add proper tests to the added code

### DIFF
--- a/src/actions/effect/EffectActions/EffectActionWithLevel.ts
+++ b/src/actions/effect/EffectActions/EffectActionWithLevel.ts
@@ -3,7 +3,7 @@ import LeveledEffectAction from "./LeveledEffectAction";
 /**
  * @class EffectActionWithLevel
  * @augments LeveledEffectAction
- * @description A class that provides a built in level() method that sets that level of the effect
+ * @description A class that provides a built in level() method that sets the level of the effect
  */
 class EffectActionWithLevel extends LeveledEffectAction {
   level(value: number | string): this {
--- a/src/actions/effect/EffectActions/EffectActionWithLevel.ts
+++ b/src/actions/effect/EffectActions/EffectActionWithLevel.ts
@@ -3,7 +3,7 @@ import LeveledEffectAction from "./LeveledEffectAction";
 /**
  * @class EffectActionWithLevel
  * @augments LeveledEffectAction
- * @description A class for all effects that include a level method
+ * @description A class that provides a built in level() method that sets that level of the effect
  */
 class EffectActionWithLevel extends LeveledEffectAction {
   level(value: number | string): this {
--- a/src/actions/effect/assistColorBlind.ts
+++ b/src/actions/effect/assistColorBlind.ts
@@ -2,9 +2,20 @@ import Action from "../Action";
 import QualifierValue from "../../qualifier/QualifierValue";
 import Qualifier from "../../qualifier/Qualifier";
 
+/**
+ * @class AssistColorBlindEffectAction
+ * @augments Action
+ *
+ * @description Applies stripes to the image to help people with common color-blind conditions to differentiate between colors that are similar for them.
+ * You can replace colors using the xray()
+ */
 class AssistColorBlindEffectAction extends Action {
   private hasStripes:boolean;
 
+  /**
+   * @@doc
+   * @description replace colors of the stripes
+   */
   xray() {
     // If stripes are defined, we don't do anything with xray
     if (!this.hasStripes) {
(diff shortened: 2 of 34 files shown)


src/actions/effect/EffectActions/EffectActionWithLevel.ts
```diff
@@ -3,7 +3,7 @@ import LeveledEffectAction from "./LeveledEffectAction";
 /**
  * @class EffectActionWithLevel
  * @augments LeveledEffectAction
- * @description A class for all effects that include a level method
+ * @description A class that provides a built in level() method that sets that level of the effect
```
strausr, line 6: I think this should be "that set the level of the effect"

src/actions/effect/assistColorBlind.ts
```diff
@@ -2,16 +2,32 @@ import Action from "../Action";
 import QualifierValue from "../../qualifier/QualifierValue";
 import Qualifier from "../../qualifier/Qualifier";
 
+/**
+ * @class AssistColorBlindEffectAction
+ * @augments Action
+ *
```
strausr, line 8: space:)

src/actions/effect/cartoonify.ts
```diff
@@ -13,12 +18,26 @@ class CartoonifyEffect extends Action {
     this.effectName = effectName;
   }
 
+  /**
+   * @description The thickness of the lines. (Range: 0 to 100, Server default: 50)
+   * @param {number} cartoonifyLevel The thickness of the lines
+   * @return {this}
+   */
   strength(cartoonifyLevel: number) {
```
strausr, line 26: Should we add return type?

src/actions/effect/leveled/loop.ts
```diff
@@ -1,6 +1,10 @@
 import LeveledEffectAction from "../EffectActions/LeveledEffectAction";
 
-
+/**
+ * @description Delivers a video or animated GIF that contains additional loops of the video/GIF.
+ * @class LoopEffectAction
+ * @augments LeveledEffectAction
```
strausr, line 6: we should have the same convention of @description at the end

src/actions/effect/outline.ts
```diff
@@ -3,6 +3,11 @@ import QualifierValue from "../../qualifier/QualifierValue";
 import Qualifier from "../../qualifier/Qualifier";
 import {prepareColor} from "../../utils/prepareColor";
 
+/**
+ * @description Adds an outline to a transparent image. For examples, see the Image Transformations guide.
```
strausr, line 7: move to end?

src/actions/effect/outline.ts
```diff
@@ -26,6 +32,7 @@ class Outline extends Action {
   /**
    * The thickness of the outline in pixels. (Range: 1 to 100, Server default: 5)
    * @param {number} width
+   * @return {this}
    */
   width(width?:number | string) {
```
strausr, line 37: return type?

strausr (changes requested):
Look good the only thing is that in certain places the description is at the beginning of the dosctring and it might be a good idea to keep the same convention and have it at the end. Also, would consider adding return type to internal functions since they were added to the jsdocs

src/actions/effect/EffectActions/EffectActionWithLevel.ts
patrick-tolosa: I'm not sure what you mean ,can you write the entire docstring?

src/actions/effect/cartoonify.ts
patrick-tolosa: added

src/actions/effect/outline.ts
patrick-tolosa: added

src/actions/effect/outline.ts
patrick-tolosa: I'm not sure, I think for functions it makes more sense for it to be first

Maybe it should always be first, even in classes?

src/actions/effect/assistColorBlind.ts
patrick-tolosa: Removed

src/actions/effect/EffectActions/EffectActionWithLevel.ts
strausr: A class that provides a built in level() method that sets the level of the effect

src/actions/effect/leveled/loop.ts
patrick-tolosa: As discussed

strausr (approved):

PR #177, opened 2020-10-06
aligned PSDTools

strausr:
#### What does this PR solve?
Aligned PSDTools
Aligned JSDocs

#### Final checklist


src/transformation/Transformation.ts
```diff
@@ -231,10 +230,11 @@ class Transformation {
   }
 
   /**
-   * @param {SmartObject} smartObject
+   * @description Adds a layer in a Photoshop document.
+   * @param action
    */
-  getSmartObject(smartObject: SmartObject): this {
-    return this.addAction(smartObject);
+  psdTools(action: IAction): this {
```
patrick-tolosa, line 236: Can this accept the right type?

it can only accept one of the PSD actions, right?

patrick-tolosa (changes requested):

PR #190, opened 2020-10-18
Aligned expression with spec

strausr:
#### What does this PR solve?
Aligned expression with spec


src/actions/variable/Variable.ts
```diff
@@ -2,6 +2,7 @@ import SetAction from "./SetAction";
 import SetReferenceAction from "./SetReferenceAction";
 import SetFromContextAction from "./SetFromContextAction";
 import SetFromMetadataAction from "./SetFromMetadataAction";
+import {ExpressionObject} from "../../values/expression/Expression";
```
patrick-tolosa, line 5: maybe a better name?
ExpressionObject is actually the class, right?

patrick-tolosa (approved):

PR #217, opened 2020-11-26
Feature/align gravity

patrick-tolosa:
#### What does this PR solve?
Completely rework how the Gravity qualifier looks and behaves.

- Aligns everything to spec (Four Gravity types: auto, ocr, compass and focusOn)
- Aligns everything to spec
- Add proper JSdocs 
- Add proper tests
- Refactor all usages in the project to fix errors after this chagne

This document requires two types of reviews:
1. Spec Review - For that you can use the tests to verify the PR is up to spec
2. Technical / Coding - Feel free to review this PR commit by commit, each commit is small and self contained - note that tests were only added after several commits.

#### Final checklist


__TESTS__/unit/values/gravity/AutoGravity.test.ts
```diff
@@ -0,0 +1,24 @@
+import {Gravity} from "../../../../src/values/gravity/Gravity";
+import {AutoFocus} from "../../../../src/values/gravity/qualifiers/autoFocus/AutoFocus";
+import {FocusOn} from "../../../../src/values/gravity/qualifiers/focusOn/FocusOn";
+
+describe('Gravity Qualifier', () => {
```
nitzanj, line 5: I didn't see a test/usage example for g_auto with weights and `avoid`, I'd add a couple of tests here (unless this is tested elsewhere and I missed it?)

src/actions/resize/ResizeActions/shared/ResizeAdvancedAction.ts
```diff
@@ -1,5 +1,5 @@
 import ResizeSimpleAction from "./ResizeSimpleAction";
-import {GravityQualifier} from "../../../../values/gravity/Gravity";
+import {IAnyGravity} from "../../../../values/gravity/GravityQualifier";
```
strausr, line 2: IAnyGravity name is a little weird why was this changed from GravityQualifier?

strausr (approved):

src/actions/resize/ResizeActions/shared/ResizeAdvancedAction.ts
patrick-tolosa: Gravity is composed of four, unrelated classes: Auto, Focus, Compass and OCR.
They are all "Gravity Qualifiers", but not all places that accept Gravity, accept *all* gravity types.

IAnyGravity is a type (not a class), that attemps to signal just that - you can pass any of the four types of Gravity.
If everything would just accept a common GravityQualifier, we would granularity over what function can accept what.

Furthermore, Gravity itself has qualifiers, which makes the terminology a bit complex. this resolves that as well.

PR #228, opened 2020-12-01
Align smart object to use multiple qualifiers

patrick-tolosa:
#### What does this PR solve?
Align smart object to use multiple qualifiers

#### Final checklist

### DIFF
--- a/__TESTS__/unit/actions/PSDTools.test.ts
+++ b/__TESTS__/unit/actions/PSDTools.test.ts
@@ -74,21 +74,21 @@ describe('Tests for Transformation Action -- PSDTools', () => {
   it('Creates a cloudinaryURL with smartObject.byIndex', () => {
     const url = new TransformableImage()
       .setConfig(CONFIG_INSTANCE)
-      .psdTools(PSDTools.smartObject().byIndex(8))
+      .psdTools(PSDTools.smartObject().byIndex(8).byIndex(5))
       .setPublicID('sample')
       .toURL();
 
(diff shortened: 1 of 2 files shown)


__TESTS__/unit/actions/PSDTools.test.ts
```diff
@@ -74,21 +74,21 @@ describe('Tests for Transformation Action -- PSDTools', () => {
...
 
-    expect(url).toBe('https://res.cloudinary.com/demo/image/upload/pg_embedded:8/sample');
+    expect(url).toBe('https://res.cloudinary.com/demo/image/upload/pg_embedded:8;5/sample');
   });
 
   it('Creates a cloudinaryURL with smartObject.byFileName', () => {
     const url = new TransformableImage()
       .setConfig(CONFIG_INSTANCE)
-      .psdTools(PSDTools.smartObject().byFileName('myFile'))
+      .psdTools(PSDTools.smartObject().byFileName('myFile').byFileName('myFile2'))
```
strausr, line 87: Not critical but why was the file name changed? is there a reason there is 2?

strausr (approved):

__TESTS__/unit/actions/PSDTools.test.ts
patrick-tolosa: It wasn't changed, it was added, to test we support two files.
i guess we could've used the same file twice, but I tried to make it clearer.

PR #232, opened 2020-12-02
Add Extract action

patrick-tolosa:
#### What does this PR solve?
Align transformation.extract()

#### Final checklist


src/actions/extract/Extract.ts
```diff
@@ -0,0 +1,68 @@
...
+class Extract extends Action {
+  private qualifierValue = new QualifierValue();
+
+  constructor() {
+    super();
+    this.qualifierValue.delimiter = ';';
+  }
+
+  /**
+   * @description Delivers an image containing only specified layers of a Photoshop image.
```
strausr, line 19: Maybe change docstring to Extracts? instead of delivers

src/actions/extract/Extract.ts
```diff
@@ -0,0 +1,68 @@
...
+   * <b>Learn more:</b> {@link https://cloudinary.com/documentation/paged_and_layered_media#deliver_selected_layers_of_a_psd_image | Deliver selected layers of a PSD image}
```
strausr, line 20: Maybe change docstring to Extracts? instead of delivers

src/transformation/Transformation.ts
```diff
@@ -254,6 +255,15 @@ class Transformation {
     return this.addAction(action);
   }
 
+  /**
+   * @doc
+   * @description Adds a page or frame from a document
```
strausr, line 260: here also would change to extracts just so naming is more clear to whoever is using

strausr (approved):

src/transformation/Transformation.ts
patrick-tolosa: Changed

src/actions/extract/Extract.ts
patrick-tolosa: Changed

PR #237, opened 2020-12-03
Align exports for Gravity

patrick-tolosa:
#### What does this PR solve?
Align exports for Gravity and fix JSDocs

#### Final checklist

### DIFF
--- a/src/qualifier/QualifierValue.ts
+++ b/src/qualifier/QualifierValue.ts
@@ -1,5 +1,8 @@
 import {Expression} from "../values/expression/Expression";
 
+/**
+ * @memberOf SDK
+ */
 class QualifierValue {
   values: any[] = [];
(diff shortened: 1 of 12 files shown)


src/qualifier/QualifierValue.ts
```diff
@@ -1,5 +1,8 @@
 import {Expression} from "../values/expression/Expression";
 
+/**
+ * @memberOf SDK
```
strausr, line 4: member of SDK?

strausr (approved):

src/qualifier/QualifierValue.ts
patrick-tolosa: Yes - I've created a new namespace for things that don't belong under "system" or "core" - these are the SDK internals (Qualifier, QualifierValue, Action etc.)

PR #251, opened 2020-12-13
Align blendmode

patrick-tolosa:
#### What does this PR solve?
Align blend mode

#### Final checklist

### DIFF
--- a/src/values/blendMode.ts
+++ b/src/values/blendMode.ts
@@ -37,16 +37,40 @@ function overlay():BlendModeQualifier {
   return new BlendModeQualifier('overlay');
 }
 
+/**
+ * @memberOf Values.BlendMode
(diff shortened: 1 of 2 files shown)


src/values/blendMode.ts
```diff
@@ -37,16 +37,40 @@ function overlay():BlendModeQualifier {
   return new BlendModeQualifier('overlay');
 }
 
+/**
+ * @memberOf Values.BlendMode
+ * @description Add an overlay image blended using the 'mask' blend mode.
```
strausr, line 42: Ads instead of Add?

strausr (approved):

src/values/blendMode.ts
patrick-tolosa: I'll leave it for the docs team,
In practice this function doesn't really add anything, it just sets the blendMode of the overlay.

PR #252, opened 2020-12-15
Add missing export for Delivery.quality

patrick-tolosa:


src/actions/delivery.ts
```diff
@@ -18,7 +18,8 @@ const Delivery = {
   density: density,
   defaultImage: defaultImage,
   colorSpace: colorSpace,
-  colorSpaceFromICC: colorSpaceFromICC
+  colorSpaceFromICC: colorSpaceFromICC,
+  quality
```
strausr, line 22: why is this not quality:quality?

strausr (approved):

src/actions/delivery.ts
patrick-tolosa: no real reason really, I'll adjust in future prs.

PR #261, opened 2020-12-18
Feature/add subtitles source

patrick-tolosa:
#### What does this PR solve?
Add Subtitles source
Fix broken StyleTransfer implementation

#### Final checklist

### DIFF
--- a/src/actions/effect/styleTransfer.ts
+++ b/src/actions/effect/styleTransfer.ts
@@ -1,9 +1,7 @@
-import {Overlay} from "../overlay";
-import {LayerAction} from "../layer/LayerAction";
-import {Qualifier} from "../../internal/qualifier/Qualifier";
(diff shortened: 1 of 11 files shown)


src/actions/effect/styleTransfer.ts
```diff
@@ -14,7 +12,7 @@ import {ImageSource} from "../../values/source/sourceTypes/ImageSource";
  */
 class StyleTransfer extends Action {
   // This action is actually a layer.
```
strausr, line 14: are these comments still relevent?

strausr (approved):

src/actions/effect/styleTransfer.ts
patrick-tolosa: Deleted, thanks!

PR #265, opened 2020-12-23
Feature/align concatenate

patrick-tolosa:
#### What does this PR solve?
Feature/align concatenate to the specification

#### Final checklist

### DIFF
--- a/src/transformation/Transformation.ts
+++ b/src/transformation/Transformation.ts
@@ -27,7 +27,7 @@ import {videoEditType} from "../actions/videoEdit";
  * @memberOf SDK
  */
 class Transformation {
-  actions: Action[];
+  actions: (Action | string)[];
 
   constructor() {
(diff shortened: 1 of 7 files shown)


src/transformation/Transformation.ts
```diff
@@ -41,6 +41,11 @@ class Transformation {
     return this;
   }
 
+  addRawAction(raw: string) {
```
strausr, line 44: this is unrelated to concatenate right?

strausr (approved):

src/transformation/Transformation.ts
patrick-tolosa: Somewhat related,
I had to include it to support this syntax:
`concatSourceTx.addRawAction(this.getTransitionString());`

We add a custom action to the source passed to concatenate.

PR #268, opened 2020-12-24
Align assist color blind to the spec

patrick-tolosa:
#### What does this PR solve?
Align assist colorblind to the spec

#### Final checklist


src/actions/effect/assistColorBlind.ts
```diff
@@ -6,10 +6,9 @@ import {Qualifier} from "../../internal/qualifier/Qualifier";
  * @augments Action
  * @description Applies stripes to the image to help people with common color-blind conditions to differentiate between colors that are similar for them.
```
strausr, line 7: I would double-check if this description is okay with docs

strausr (approved):

PR #282, opened 2020-12-29
Various alignments to the code generator

patrick-tolosa:
#### What does this PR solve?
Various alignments to the code generator

- Change case (CMYK to cmyk)
- Add safety checks to prepareColor if color is not passed
- Change pixelWidth to squareSize for pixelate
- add tests(and support) for roundCorners radius with value 0
- Add prepareColor to text and subtitle layers' colors
- Add an explicit test for the base64 encoding for remote functions

#### Final checklist

### DIFF
--- a/src/actions/effect.ts
+++ b/src/actions/effect.ts
@@ -19,6 +19,8 @@ import {DitherEffectAction} from "./effect/dither";
 import {DeshakeEffectAction} from "./effect/leveled/deshake";
 import {Pixelate} from "./effect/pixelate/pixelate";
 import {ImageSource} from "../values/source/sourceTypes/ImageSource";
+import {EffectActionWithStrength} from "./effect/EffectActions/EffectActionWithStrength";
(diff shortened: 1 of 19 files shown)


src/actions/effect.ts
```diff
@@ -82,8 +84,8 @@ function colorize(colorizeLevel?: number):ColorizeEffectAction {
  * @param {number} oilPaintLevel The strength of the effect. (Range: 0 to 100, Server default: 30)
  * @return {EffectActionWithLevel}
```
strausr, line 85: update return

src/actions/effect.ts
```diff
@@ -161,8 +163,8 @@ function advancedRedEye():SimpleEffectAction {
  * @param {number | string} level The balance between black (100) and white (0). (Range: 0 to 100, Server default: 50)
  * @return EffectActionWithLevel
```
strausr, line 164: update return type

PR #285, opened 2020-12-31
Add support for raw action and transformation

patrick-tolosa:
#### What does this PR solve?
Add support for raw action and transformation
- When a string is passed to `addAction()`, validate it does not contain a forward slash
- When a string is passed to `addTransformation()`, perform no validation.

#### Final checklist


src/assets/CloudinaryTransformable.ts
```diff
@@ -215,10 +215,10 @@ class CloudinaryTransformable extends CloudinaryFile {
 
   /**
    * @desc A proxy to {@link SDK.Transformation| Transformation} - Calls the same method contains in this.transformation
```
strausr, line 217: update docs?

strausr (approved):

PR #292, opened 2021-01-04
Feature/small spec and codegen alignments

patrick-tolosa:
#### What does this PR solve?
See commit list in the PR for tasks done. (each commit is isolated)

#### Final checklist

### DIFF
--- a/src/actions/effect/vectorize.ts
+++ b/src/actions/effect/vectorize.ts
@@ -18,17 +18,17 @@ class VectorizeEffectAction extends Action {
 
   /**
    * @description The number of colors. (Range: 2 to 30, Server default: 10)
-   * @param {number} num
+   * @param {number | string} num
    * @return {this}
    */
-  numOfColors(num: number): this {
+  numOfColors(num: number | string): this {
     this._numOfColors = num;
     return this;
   }
 
   /**
    * @description The level of detail. Specify either a percentage of the original image (Range: 0.0 to 1.0) or an absolute number of pixels (Range: 0 to 1000). (Server default: 300)
-   * @param {number} num
+   * @param {number | string} num
    * @return {this}
    */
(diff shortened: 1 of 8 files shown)


src/actions/effect/vectorize.ts
```diff
@@ -31,7 +31,7 @@ class VectorizeEffectAction extends Action {
    * @param {number} num
```
strausr, line 31: add string

src/actions/effect/vectorize.ts
```diff
@@ -41,7 +41,7 @@ class VectorizeEffectAction extends Action {
...
    * @return {this}
    */
-  despeckleLevel(num: number):this {
+  despeckleLevel(num: number | string):this {
```
strausr, line 44: add string

src/assets/CloudinaryFile.ts
```diff
@@ -57,6 +57,9 @@ class CloudinaryFile {
    * @return {string} CloudinaryURL
    */
   createCloudinaryURL(transformation?: Transformation): string {
+    if (typeof this.cloudName === 'undefined') {
+      throw 'You must supply a cloudName in either toURL() or whe initializing the asset';
```
strausr, line 61: I think there is a typo here?

strausr (approved):

src/assets/CloudinaryFile.ts
patrick-tolosa: yep, fixed.

src/actions/effect/vectorize.ts
patrick-tolosa: added

src/actions/effect/vectorize.ts
patrick-tolosa: added

PR #296, opened 2021-01-07
Feature/align video codec

patrick-tolosa:
#### What does this PR solve?
Add video codec, with level and profile
Fix namespaces in JSDoc to properly show links for VideoCodec
Fix a wide-spread incorrect use of the @Augment keyword

#### Final checklist

### DIFF
--- a/src/values/videoCodecLevel.ts
+++ b/src/values/videoCodecLevel.ts
@@ -1,0 +1,67 @@
+/**
+ * @description Something Something Dark Side
+ * @memberOf Values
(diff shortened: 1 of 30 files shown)


src/values/videoCodecLevel.ts
```diff
@@ -0,0 +1,67 @@
+/**
+ * @description Something Something Dark Side
```
strausr, line 2: Fix:)

strausr (approved):

PR #309, opened 2021-01-17
change addTransformation() to accept a transformation instance as well as a string

patrick-tolosa:
#### What does this PR solve?
change addTransformation() to accept a transformation instance as well as a string

#### Final checklist

### DIFF
--- a/src/assets/CloudinaryTransformable.ts
+++ b/src/assets/CloudinaryTransformable.ts
@@ -224,6 +224,15 @@ class CloudinaryTransformable extends CloudinaryFile {
   }
 
   /**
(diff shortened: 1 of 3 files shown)


src/assets/CloudinaryTransformable.ts
```diff
@@ -223,6 +223,15 @@ class CloudinaryTransformable extends CloudinaryFile {
     return this;
   }
 
+  /**
+   * @description Extend your transformation with another transformation
```
strausr, line 227: should we change the name to extendTransformation rather than add?

strausr (approved):

PR #317, opened 2021-01-20
Fix missing @docs and removed an incorrect qualifier

patrick-tolosa:
#### What does this PR solve?
Fix missing @docs and removed an incorrect qualifier
(removed the AutoCodec.auto qualifier which isn't specced or documented)

#### Final checklist

### DIFF
--- a/src/values/chromaSubSampling.ts
+++ b/src/values/chromaSubSampling.ts
@@ -1,18 +1,17 @@
 
 /**
- * @@doc
  * @memberOf Values.ChromeSubSampling
  */
 function chroma444():number { return 444; }
 
 /**
- * * @@doc
  * @memberOf Values.ChromeSubSampling
  */
 function chroma420():number { return 420; }
 
 
 /**
+ * @docs
  * @memberOf Values
  * @namespace ChromeSubSampling
  */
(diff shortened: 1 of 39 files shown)


src/values/videoCodecLevel.ts
```diff
@@ -7,60 +7,52 @@
 
 
 /**
- * @@doc
  * @memberOf Values.VideoCodecLevel
  * @return {string}
```
strausr, line 11: Are these not returning numbers?

strausr (approved):

src/values/chromaSubSampling.ts
```diff
@@ -1,18 +1,17 @@
...
 
 /**
- * * @@doc
  * @memberOf Values.ChromeSubSampling
  */
 function chroma420():number { return 420; }
 
 
 /**
+ * @docs
```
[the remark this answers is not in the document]
patrick-tolosa: yes, a second PR will be made after this one.

src/values/videoCodecLevel.ts
patrick-tolosa: they are :) fixed.

PR #327, opened 2021-01-25
Add relevant overlay tests from cloudinary-core

maoznir:
Add relevant overlay tests from cloudinary-core:
Some of the added tests fail and so we should consider whether they are indeed needed or update the spec/implementation.
Failing tests:
1. 'should serialize a text source'
Not converting spaces to %20

2. 'should serialize a unicode url of fetch source'
Breaking on chinese characters

3. 'should support string interpolation'
This is also failing due to not converting spaces to %20

4. 'should throw an exception if fontFamily is not provided'
TextStyle(null, 17)


patrick-tolosa:
@maoznir can you please add `it.skip(...)` to the failing tests so it's easier to see wha's failing?

Also, can you add a short comment as to what's failing? 
I'm guessing the variable encoding within overlays fail, but I'm not sure we need that moving forward.

patrick-tolosa (commented):
@maoznir can you please add it.skip(...) to the failing tests so it's easier to see wha's failing?

Also, can you add a short comment as to what's failing?
I'm guessing the variable encoding within overlays fail, but I'm not sure we need that moving forward.

patrick-tolosa (approved):

PR #334, opened 2021-01-28
Fix an incorrect test and a missing export from index

patrick-tolosa:
#### What does this PR solve?
Fix an incorrect test and a missing export from index

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -35,5 +35,6 @@ export {
   CloudinaryImage,
   CloudinaryVideo,
   CloudinaryFile,
+  CloudinarySDK,
   CloudinarySDK as default
 };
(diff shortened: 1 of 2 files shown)


src/index.ts
```diff
@@ -35,5 +35,6 @@ export {
   CloudinaryImage,
   CloudinaryVideo,
   CloudinaryFile,
+  CloudinarySDK,
```
strausr, line 38: twice?

strausr (approved):

src/index.ts
patrick-tolosa: one is a default export. 

import {CloudinarySDK} from '...'
import CloudinarySDK from ...

PR #335, opened 2021-01-28
Set the bundle target to ES6 for ESM, and ES5 for UMD

patrick-tolosa:
#### What does this PR solve?
- Set the bundle target to ES6 for ESM, and ES5 for UMD
- Dropped babel (we weren't using it anyway)


patrick-tolosa:
The PR fails due to code coverage, but the other open PRs should fix that, so this can be merged last.

strausr (approved):

PR #346, opened 2021-02-07
Add examples to IURlConfig interfaces

patrick-tolosa:
#### What does this PR solve?
- Hide private interfaces that aren't needed yet (they'll be needed for node)
- Add examples to existing members
- Fixed a small type error in the docs

### DIFF
--- a/src/config/interfaces/Config/IURLConfig.ts
+++ b/src/config/interfaces/Config/IURLConfig.ts
@@ -2,8 +2,9 @@
  * @name ICloudinaryAssetConfigurations
  * @summary config
  * @description Defines the configuration needed for URL-related options when creating Cloudinary URL
+ * <b>Learn more:</b> {@link https://cloudinary.com/documentation/cloudinary_sdks#configuration_parameters|URL Parameters}
  * @prop {string}  [cname]
- * @prop {boolean} [secureDistribution]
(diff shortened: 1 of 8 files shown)


strausr (approved):

src/config/interfaces/Config/IURLConfig.ts
```diff
@@ -54,10 +69,29 @@ interface IURLConfig {
...
   shorten?: boolean;
+
+  /**
+   * Whether or not to use the root path
+   */
   useRootPath?: boolean;
+
+  /**
+   * Whether or not to force a version
+   */
```
carlevison, line 94: I think that some of these need more explanation. Maybe just some examples would help. What does it mean to shorten the URL? What's the root path? What does it mean to force a version?

patrick-tolosa: I've linked this whole file (at the top) to the config params with a learn more to avoid adding mountains of texts to explain each and every config.

Consider how complicated it is to expalin the difference between `secureDistribution`, `secure`, and `cname`.

patrick-tolosa: @carlevison I added an internal ticket to extend these in time, once we finish with other priorities

PR #358, opened 2021-02-28
Change the public property of an asset to private

patrick-tolosa:
#### What does this PR solve?

This PR is based on #357 (The build currently fails due to a compilation error)

All of the public properties have setters.
These properties should not be 'read' after set (or at least it's not specced that way), this PR removes the public properties by changing them to private.


src/assets/CloudinaryFile.ts
```diff
@@ -23,11 +23,12 @@ class CloudinaryFile {
...
-  public publicID: string;
-  public extension: string;
-  public suffix: string;
-  public storageType: string; // type upload/private
+  private version: number | string;
+  private publicID: string;
+  private extension: string;
+  private signature: string;
+  private suffix: string;
+  private storageType: string; // type upload/private
```
strausr, line 31: update comment

strausr (approved):

PR #378, opened 2021-04-08
Add new setters to variable: float, integer and string

patrick-tolosa:
#### What does this PR solve?
Add new setters to variable: float, integer and string

1. Modify setAction so that the wrapper of a string can be controlled (We needed to control the presence of the ! wrapper)
2. Add three setter methods in Variable
3. Each method performs the required parsing and passes the parsed result into the original SetAction constructor

### DIFF
--- a/src/actions/variable.ts
+++ b/src/actions/variable.ts
@@ -90,9 +90,7 @@ function setInteger(name: string, value: number): SetAction {
  * @return {Actions.Variable.SetAction}
  */
 function setString(name: string, value: string | number): SetAction {
-  const val = typeof value === 'string' ? value : value.toString();
-
-  return new SetAction(name, val);
+  return new SetAction(name, value.toString());
 }
 
 
--- a/src/actions/variable.ts
+++ b/src/actions/variable.ts
@@ -3,6 +3,7 @@ import SetAssetReferenceAction from "./variable/SetAssetReferenceAction";
 import SetFromContextAction from "./variable/SetFromContextAction";
 import SetFromMetadataAction from "./variable/SetFromMetadataAction";
 import {ExpressionQualifier} from "../qualifiers/expression/ExpressionQualifier";
+import {toFloatAsString} from "../internal/utils/toFloatAsString";
 
 /**
  * Defines a new user variable with the given value.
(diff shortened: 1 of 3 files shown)


src/actions/variable.ts
```diff
@@ -34,14 +35,69 @@ import {ExpressionQualifier} from "../qualifiers/expression/ExpressionQualifier"
...
+
+/**
+ * @summary action
+ * @description Same as 'set', but forces the end value to be a string setString(1) will result in $foo_!1!
+ * @memberOf Actions.Variable
+ * @param {string, number} name Variable name
+ * @param {number} value Variable value
+ * @return {Actions.Variable.SetAction}
+ */
+function setString(name: string, value: string | number): SetAction {
```
maoznir, line 92: Why allow number?

src/actions/variable.ts
```diff
@@ -34,14 +35,69 @@ import {ExpressionQualifier} from "../qualifiers/expression/ExpressionQualifier"
...
+  const val = typeof value === 'string' ? value : value.toString();
```
maoznir, line 93: value.toString will work for any of the allowed types, no? so can remove the check.

src/actions/variable.ts
```diff
@@ -34,14 +35,69 @@ import {ExpressionQualifier} from "../qualifiers/expression/ExpressionQualifier"
...
+
+  return new SetAction(name, val);
```
maoznir, line 95: Maybe rewrite like this:
return new SetAction(name, value.toString());

maoznir (changes requested):

src/actions/variable.ts
patrick-tolosa: I thought it might be a good idea to support it as well, but we can make it strictly a string.

In that case I wonder if we really need the `setString()` function, as `set('foo')` and `setString('foo')` do the same thing and are not ambigious.

What do you think?

src/actions/variable.ts
patrick-tolosa: Done

src/actions/variable.ts
patrick-tolosa: Done

src/actions/variable.ts
maoznir: So lets leave it like you did

maoznir (approved):

PR #383, opened 2021-04-25
Add failing tests that cover more backward compatibility cases

yoadsn:
#### What does this PR solve?
Highlights some backward compatibility failed-cases. (adds the test cases).

Here are the failing cases:
- "radius" does not accept a string (treats as null) or an array (crashes) - Bug is around [here](https://github.com/cloudinary/cloudinary-js-base/blob/365d9254736ef6d6c7ad0ef216a4d85ff4683401/src/backwards/transformationProcessing/processRadius.ts#L17).
- "border" should accept an object not just a string (V1 accepts an object see [here](https://github.com/cloudinary/cloudinary_js/blob/8859ceb9ab864903e820d1eef2287897808379fe/src/transformation.js#L592))
- transformation "layers" should clone before modifying "layer" options (See [here](https://github.com/cloudinary/cloudinary-js-base/blob/365d9254736ef6d6c7ad0ef216a4d85ff4683401/src/backwards/generateTransformationString.ts#L100))

### DIFF
--- a/__TESTS__/backwardsComaptibility/createV1URL.test.ts
+++ b/__TESTS__/backwardsComaptibility/createV1URL.test.ts
@@ -81,6 +81,23 @@ describe('Create v1 urls', () => {
     expect(url).toEqual("https://res.cloudinary.com/test123/image/upload/if_w_lt_200,c_fill,h_120,w_80/if_w_gt_400,c_fit,h_150,w_150/e_sepia/sample");
   });
 
+  it("should not modify options objects when chaining transformations ", function () {
+    const optionObj = {
+      cloud_name: 'test123',
+      transformation: [{
+        overlay: 'somepid',
+        crop: "fill",
+        height: 120,
+        width: 80
+      }]
+    };
+    const urlFirstRun = createCloudinaryV1URL("sample", optionObj);
+    const urlSecondRun = createCloudinaryV1URL("sample", optionObj);
+
+    expect(urlFirstRun).toEqual("https://res.cloudinary.com/test123/image/upload/c_fill,h_120,w_80/sample");
+    expect(urlFirstRun).toEqual(urlSecondRun);
+  });
+
   it("should translate operators", function () {
     const url = createCloudinaryV1URL("sample", {
       cloud_name: 'test123',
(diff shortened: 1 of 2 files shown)


__TESTS__/backwardsComaptibility/createV1URL.test.ts
```diff
@@ -81,6 +81,23 @@ describe('Create v1 urls', () => {
...
+    const optionObj = {
+      cloud_name: 'test123',
+      transformation: [{
+        overlay: 'somepid',
+        crop: "fill",
+        height: 120,
+        width: 80
+      }]
+    };
+    const urlFirstRun = createCloudinaryV1URL("sample", optionObj);
```
patrick-tolosa, line 94: This behaviour should match the v1 behaviour, where the options are mutated as they're being consumed by the API.

We'll confirm if there's a regression from v1 and we'll deploy a fix if needed.

yoadsn: I see, that makes sense since I used the same options objects in V1 and V2 one after the other this highlighted that point. Indeed this is not the normal use case to "parse" the same options object twice.

How do you feel about adding V1 as a dev dependency and using it in tests to get a better compatibility test coverage?

This way one could build a simple infrastructure that does (pseudo code):
```
(pubid, options, config) => expect(
    v2url(pubid, {...config, ...options})
  ).toEqual(
    v1url(config)(pubid, options)
)
```
And adding the test cases would be a breeze - covering 100's of cases with very little effort.

patrick-tolosa:
Closing this PR, as the issues raised here we merged in #386 #384 

We'll probably release these next week with a few more minor changes.

yoadsn:
Thanks @patrick-tolosa!
Have you considered my suggestion to include tests that use the v1 SDK and compare it to the results of the V2 SDK to have a verbatim comparison? Would you like me to try and create such a PR to extend the tests?

patrick-tolosa:
The V2 SDK should not create the exact URL as in V1, however the backwards compatibility method should do that.

Having a PR that proves that would indeed be useful.

PR #391, opened 2021-05-26
Add back comp tests against the real V1 SDK

yoadsn:
This does not implement all test cases by all means. It shows how we could use test cases taken from [here](https://github.com/cloudinary/cloudinary_js/blob/master/test/spec/automatic/transformation-spec.js) and reuse them as part of the backward compatibility spec testing of a V1 URL generation.

#### What does this PR solve?
This PR attempts to create a testing infrastructure to allow faster and more accurate backward compatibility testing for URL generation against the real V1 SDK. If used correctly - can greatly improve the coverage.

#### Final checklist


yoadsn:
The idea here is that we can simply copy "options" objects from samples and spec tests or bur repros, put them inside the "cases" array with a proper name and get another coverage boost tested against the real V1 sdk which can be updated from time to time to ensure compatibility is carried on.

@patrick-tolosa Please note that one tests that I marked as skip is not passing - i.e It shows a mismatch between the v1 url generation logic of the two sdks.

yoadsn:
@patrick-tolosa I notice now the #390 "v1" -> "legacy" PR - once you complete it to master I will pull from it, merge the conflicts and fix my naming to use the "legacy" term.

patrick-tolosa:
This looks really good,
The failing test you have is unfortunately coming from an inconsistency within the old SDKs:

In the NodeJS SDK the skipped test will fail, while in the `cloudinary-core` package, the test will pass.

We've yet to decide how to address that.

yoadsn:
@patrick-tolosa Lets keep it skipped then.
LMK if you generally are ok with the PR - I can wait until you complete #390, align this PR and ask for your review again.
Thanks!

patrick-tolosa:
Sounds good @yoadsn

patrick-tolosa:
@yoadsn FYI we merged #390

yoadsn:
@patrick-tolosa FYI - I synced with master, adapted V1->Legacy naming, added tests, improved test reliability by cloning the options before using them against the legacy API lib.

Also - I have added 2 skipped failing tests around the "format" options.
"should not repeat extension"
"options format should override file extension"
Where basically the old SDK would not disregard a public Id file extension when the "format" option was specified.

patrick-tolosa (approved):

patrick-tolosa:
Thanks @yoadsn  - Merged!

PR #407, opened 2021-07-29
Add 'p_' to code generation unsupported list

maoznir:


patrick-tolosa:
Closing we as we discussed this should be handled in the code generation service itself and not partially in sdks.

PR #417, opened 2021-08-04
Improve factory types for Resize

patrick-tolosa:
This PR is for discussion purposes.

Currently, our actions provide unintuitive hints in the IDE, for example resize() accepts a ResizeAction type, but in fact we want our users to use the factory methods (crop(), scale() etc.)

This PR offers a typescript solution that both improves the type safety (somewhat), and improves IDE hints

```
img.resize(new ResizeAction()) // Will now be a type error
img.resize(scale()) // Will work as expected
```

### Concerns
- We are now casting the result of our actions `as ...` so we are losing type safety in our factory methods, this should be mitigated through proper tests, but it's still a concern.

This PR updates both a test file and the main transformation file.

### DIFF
--- a/src/actions/resize.ts
+++ b/src/actions/resize.ts
@@ -70,8 +70,8 @@ import {ResizeFillAction} from "./resize/ResizeFillAction";
  * @param {number|string} height The required height of a transformed asset.
  * @return {Actions.Resize.ScaleAction}
  */
-function scale(width?: number | string, height?: number | string): ScaleAction {
-  return new ScaleAction('scale', width, height);
+function scale(width?: number | string, height?: number | string): { __EXTENDED__:true } & ScaleAction {
+  return new ScaleAction('scale', width, height) as ReturnType<typeof scale>;
 }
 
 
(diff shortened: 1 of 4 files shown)


src/actions/resize.ts
```diff
@@ -70,8 +70,8 @@ import {ResizeFillAction} from "./resize/ResizeFillAction";
  * @param {number|string} height The required height of a transformed asset.
  * @return {Actions.Resize.ScaleAction}
  */
-function scale(width?: number | string, height?: number | string): ScaleAction {
-  return new ScaleAction('scale', width, height);
+function scale(width?: number | string, height?: number | string): { __EXTENDED__:true } & ScaleAction {
```
maoznir, line 73: what is __EXTENDED__ ?

patrick-tolosa: Just a private key (can be named anything), to ensure that this return type is NOT just new ScaleAction.

This ensures that `resize(new ScaleAction())` is a type error, while `resize(scale())` is not a type error

maoznir: ok i see, how does the hint look like in the ide after this change?

patrick-tolosa: Good in some areas, bad in others.

In typescript, the IDE hints are amazing, but the type errors are shown in a very bad way. 
This doesn't look like a very good solution unfortunately.

patrick-tolosa:
Irrelevant for now

PR #429, opened 2021-09-02
Fix lodash.clonedeep module issue and add module key to package.json

patrick-tolosa:
- Copy lodash.clonedeep to a file instead of importing it
- Add a module key to package.json
- Fix various linting issues (Delete unused lines, add missing return types)

#### What does this PR solve?
... A few words

#### Final checklist


src/config/CloudinaryConfig.ts
```diff
@@ -33,16 +31,6 @@ class CloudinaryConfig {
     return this;
   }
 
-  setAPIConfig(apiConfig: IApiConfig): this {
-    // placeholder
```
patrick-tolosa, line 37: This was added as a placeholder but was never used.

src/package.json
```diff
@@ -11,5 +11,6 @@
 	},
 	"main": "./bundles/umd/base.js",
 	"browser": "./index.js",
+	"module": "./index.js",
```
patrick-tolosa, line 14: Module is used by some build tools instead of main to correctly infer where the ESM code lives.

src/internal/utils/cloneDeep.ts
```diff
@@ -1,11 +1,1767 @@
...
+ * _.isObjectLike([1, 2, 3]);
+ * // => true
+ *
+ * _.isObjectLike(_.noop);
+ * // => false
+ *
+ * _.isObjectLike(null);
+ * // => false
+ */
+function isObjectLike(value) {
```
maoznir, line 1692: Seeing as we already have an isObject function, maybe we should move such functions to their own file and import here.

maoznir (approved):

src/internal/utils/cloneDeep.ts
patrick-tolosa: We could but I wouldn't change this specific file, as it's an exact copy of lodash (we could potentially introduce new bugs by changing the implementation.

what do you think?

strausr (approved):

PR #433, opened 2021-09-09
Add package.exports

maoznir:
This pr adds an auto generated exports property to package.json
This solves the issues with importing url-gen for the default svelte app configuration


package.json
```diff
@@ -88,5 +88,34 @@
...
+    "./actions/reshape": "./actions/reshape.js",
+    "./actions/resize": "./actions/resize.js",
+    "./actions/rotate": "./actions/rotate.js",
+    "./actions/roundCorners": "./actions/roundCorners.js",
+    "./actions/transcode": "./actions/transcode.js",
+    "./actions/variable": "./actions/variable.js",
+    "./actions/videoEdit": "./actions/videoEdit.js",
+    "./qualifiers/animatedFormat": "./qualifiers/animatedFormat.js",
+    "./qualifiers/aspectRatio": "./qualifiers/aspectRatio.js",
+    "./qualifiers/background": "./qualifiers/background.js",
```
patrick-tolosa, line 110: It looks like we can export entire folders instead of manually listing these one by one:

https://nodejs.org/api/packages.html

```
{
  "name": "my-mod",
  "exports": {
    ".": "./lib/index.js",
    "./lib": "./lib/index.js",
    "./lib/*": "./lib/*.js",
    "./feature": "./feature/index.js",
    "./feature/*": "./feature/*.js",
    "./package.json": "./package.json"
  }
}
```

package.json
```diff
@@ -88,5 +88,34 @@
   "dependencies": {
     "@types/lodash.clonedeep": "^4.5.6",
     "lodash.clonedeep": "^4.5.0"
+  },
+  "exports": {
+    ".": "./index.js",
+    "./package.json": "./package.json",
+    "./actions/adjust": "./actions/adjust.js",
+    "./actions/customFunction": "./actions/customFunction.js",
```
patrick-tolosa, line 96: We're missing the entry point for the UMD bundle

`
Warning: Introducing the "exports" field prevents consumers of a package from using any entry points that are not defined, including the package.json (e.g. require('your-package/package.json'). This will likely be a breaking change.
`

looks like we need to specify *all* entrypoints.

package.json
```diff
@@ -88,5 +88,34 @@
   "dependencies": {
     "@types/lodash.clonedeep": "^4.5.6",
     "lodash.clonedeep": "^4.5.0"
+  },
+  "exports": {
```
patrick-tolosa, line 92: Exports require special treatment for mixing ESM and CommonJS (Our bundle package is UMD for example)

`
For more information about supporting both CommonJS and ES Modules in a single package please consult the dual CommonJS/ES module packages section.
`

Did we test to ensure that still works?

patrick-tolosa (changes requested):

package.json
maoznir: done

package.json
maoznir: tested on svelte + svelte ssr and updated "." value accordingly, please test on more frameworks

strausr (approved):

patrick-tolosa (approved):

PR #434, opened 2021-09-09
Inject package version during build instead of importing it from pack…

maoznir:
This pr fixes end users errors related to importing from package.json file.
The package version is injected during build instead of importing it from package.json.
Also mocked the version value during unit tests, because they rely on source code and not on build result.
I also tested manually by building and looking for the value in the resulted bundle, it was indeed replaced during build.


rollup.config.js
```diff
@@ -15,6 +17,9 @@ export default [{
   plugins: [
     json(),
     resolve(),
+    replace({
+      'PACKAGE_VERSION_INJECTED_BY_ROLLUP_DURING_BUILD': version
```
patrick-tolosa, line 21: This correctly addresses the `rollup` build process, however we're also using `tsc` to create the ESM files.

The value `PACKAGE_VERSION_INJECTED_BY_ROLLUP_DURING_BUILD` does not get replaced on build time in this file `/js-url-gen/dist/internal/utils/packageVersion.js`, the tests miss this because the file is mocked.

Maybe we can run add a new test that runs on deployment that builds the project first, then tests on the `./dist` file?

patrick-tolosa (changes requested):

rollup.config.js
maoznir: Fixed with replaceStrings.ts

scripts/replaceStrings.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint @typescript-eslint/no-var-requires: 0 */
```
strausr, line 1: The name of this file is a bit confusing. Maybe we should rename it to something like replacePakcageVersion 
setPackageVersion

strausr (changes requested):

scripts/replaceStrings.ts
maoznir: updated file name

patrick-tolosa (approved):

strausr (approved):

PR #441, opened 2021-09-19
Adjust the error when the cloudName is missing

patrick-tolosa:
- we cannot at this point set this value through the toURL() method


patrick-tolosa:
It looks like we had failing lint rules in master, possibly due to multiple merges,

In any case, this PR also fixes those.

strausr (approved):

PR #455, opened 2021-10-21
Fix travis.yml to only run tests for pr and not also for branch

ghost:
This pr prevents travis from checking both the pr (merged to master) and the branch (without merging to master)
This will reduce the travis resource usage by 50% and will prevent the need to wait for an unneeded test.


maoznir:
Closing as this will prevent all tests for remote branches.

PR #483, opened 2021-11-10
Disable analytics for publicId with question mark

maoznir:


strausr (approved):

const-cloudinary:
@maoznir , in the code it would be better to include a comment, saying why we exclude analytics for public IDs with `?`. Nobody will remember it in 6 months or so.

PR #498, opened 2022-01-11
Add ConcatenateAction.fromJson()

maoznir:
This pr adds fromJson support to Concatenate.
It will be used as base for adding fromJson support to more video.edit actions.

Reviewer please note: I've added a new parameter to the internal fromJson function: tranformationFromJson.
It is used by actions that have internal transformations (like Concatenate) to generate a transformation from json.
Adding the parameter prevents the need to import the function and all it's dependencies inside the Concatenate action or it's source qualifier.

### DIFF
--- a/src/actions/videoEdit/ConcatenateAction.ts
+++ b/src/actions/videoEdit/ConcatenateAction.ts
@@ -3,6 +3,9 @@ import {Transformation} from "../../transformation/Transformation.js";
 import {VideoSource} from "../../qualifiers/source/sourceTypes/VideoSource.js";
 import {ImageSource} from "../../qualifiers/source/sourceTypes/ImageSource.js";
 import {FetchSource} from "../../qualifiers/source/sourceTypes/FetchSource.js";
+import {IActionModel} from "../../internal/models/IActionModel.js";
+import {IConcatenateActionModel} from "../../internal/models/IConcatenateActionModel.js";
+import {ITransformationFromJson} from "../../internal/models/IHasFromJson.js";
 
 /**
  * @description Class for Concatenating another video.
@@ -17,6 +20,7 @@ class ConcatenateAction extends Action {
   private _prepend: boolean;
   private _duration: number;
   private _transition: VideoSource;
+  protected _actionModel: IConcatenateActionModel;
 
   /**
    *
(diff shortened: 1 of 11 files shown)


src/actions/videoEdit/ConcatenateAction.ts
```diff
@@ -25,6 +29,11 @@ class ConcatenateAction extends Action {
    */
   constructor(source: VideoSource | ImageSource | FetchSource) {
     super();
+    this._actionModel = {
+      actionType: 'concatenate',
+      source: { sourceType: 'video' }
```
maoznir, line 34: When we add toJson support, we'll use it to update the source model

src/actions/videoEdit/ConcatenateAction.ts
```diff
@@ -124,6 +133,28 @@ class ConcatenateAction extends Action {
...
     ].filter((a) => a).join('/');
   }
+
+  static fromJson(actionModel: IActionModel, transformationFromJson: ITransformationFromJson): ConcatenateAction {
+    const {source, transition, prepend, duration} = (actionModel as IConcatenateActionModel);
+
+    // We are using this() to allow inheriting classes to use super.fromJson.apply(this, [actionModel])
+    // This allows the inheriting classes to determine the class to be created
+    // @ts-ignore
+    const result = new this(VideoSource.fromJson(source, transformationFromJson));
```
patrick-tolosa, line 143: For clarity, can we replace `new this()` with `new ConcatenateAction`?

I'm pretty sure nothing extends `ConcatenateAction` (so we don't need to have a very generic solution.

I also don't understand the comment, what inheriting classes are you referring to here?

src/qualifiers/source/sourceTypes/VideoSource.ts
```diff
@@ -35,6 +37,21 @@ class VideoSource extends BaseSource {
...
   }
+
+  static fromJson(qualifierModel: IVideoSourceModel, transformationFromJson: ITransformationFromJson): VideoSource {
+    const videoSourceModel = qualifierModel as IVideoSourceModel;
+    const {publicId, transformation} = videoSourceModel;
+
+    // We are using this() to allow inheriting classes to use super.fromJson.apply(this, [qualifierModel])
+    // This allows the inheriting classes to determine the class to be created
+    // @ts-ignore
+    const result = new this(publicId);
```
patrick-tolosa, line 48: Same comment as above

src/actions/videoEdit/ConcatenateAction.ts
maoznir: we have a "new this" in every fromJson, the comment is copy pasted in all of them for any newcomers to understand why we chose to do that. even if nothing extends it now, it might be extended. also, this technique is used by some inheriting resize actions for example, so we keep it aligned.

PR #501, opened 2022-01-12
Feature/concatenate to json

maoznir:
* Add ConcatenateAction.toJson()
* Add QualifierModel & ActionModel classes that can be extended by any class that needs a model + toJson

### DIFF
--- a/__TESTS__/unit/toJson/concatenate.toJson.test.ts
+++ b/__TESTS__/unit/toJson/concatenate.toJson.test.ts
@@ -1,0 +1,32 @@
+import {Transformation} from '../../../src';
+import {Concatenate} from "../../../src/qualifiers/concatenate";
+import ConcatenateAction from "../../../src/actions/videoEdit/ConcatenateAction";
+
+describe('concatenate.toJson()', () => {
+  it('Concatenate + prepend + duration + transition', () => {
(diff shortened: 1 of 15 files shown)


__TESTS__/unit/toJson/concatenate.toJson.test.ts
```diff
@@ -0,0 +1,32 @@
+import {Transformation} from '../../../src';
+import {Concatenate} from "../../../src/qualifiers/concatenate";
+import ConcatenateAction from "../../../src/actions/videoEdit/ConcatenateAction";
+
+describe('concatenate.toJson()', () => {
+  it('Concatenate + prepend + duration + transition', () => {
+    const transformation = new Transformation()
+      .videoEdit(
+        new ConcatenateAction(Concatenate.videoSource('dog'))
```
patrick-tolosa, line 9: We should use `VideoEdit.concatenate()` instead of `new Concatenate`, if it doesn't exist we should create it.

patrick-tolosa (approved):

__TESTS__/unit/toJson/concatenate.toJson.test.ts
maoznir: done

PR #510, opened 2022-02-03
Fix jsDoc generation error by excluding type guards

maoznir:
This pr fixes the build on Node 16 by moving type guards to /typeGuards & excluding that folder in jsdoc.config.js
jsDoc gives errors for typescript type guards, example:

```
ERROR: Unable to parse a tag's type expression for source file
Invalid type expression "obj is IActionModel": Expected "|" but "i" found.
```


patrick-tolosa:
I don't feel comfortable fixing it like this, splitting the codebase should not be a decision dictated by tools.
Do we know why this fails for Node 16?

maoznir:
@patrick-tolosa I could not find the reason for the node 16 issues, I looked in the jsDoc repo and didn't find answers there too, I saw similar issues reported, like this one: https://github.com/jsdoc/jsdoc/issues/1961

patrick-tolosa:
Lets stick with Node 14 then, this doesn't feel right, and we should address the Node 16 issue.

maoznir:
closing this, we've locked to a specific jsDocs version that does not have this error. We will upgrade jsDoc if needed after it's issue with ts guards is fixed

PR #518, opened 2022-02-16
fix makeTransparent

strausr:
#### What does this PR solve?
fix makeTransparent fromJson to return e_make_transparent rather than e_makeTransparent

### DIFF
--- a/src/actions/effect/leveled/MakeTransparent.ts
+++ b/src/actions/effect/leveled/MakeTransparent.ts
@@ -5,6 +5,7 @@ import {prepareColor} from "../../../internal/utils/prepareColor.js";
 import {SystemColors} from "../../../qualifiers/color.js";
 import {IMakeTransparentEffectModel} from "../../../internal/models/IEffectActionModel.js";
 import {IActionModel} from "../../../internal/models/IActionModel.js";
+import {ACTION_TYPE_TO_EFFECT_MODE_MAP} from "../../../internal/internalConstants.js";
 
 /**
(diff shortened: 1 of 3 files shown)


src/actions/effect/leveled/MakeTransparent.ts
```diff
@@ -40,7 +41,7 @@ class MakeTransparentEffectAction extends LeveledEffectAction {
 
     // We are using this() to allow inheriting classes to use super.fromJson.apply(this, [actionModel])
     // This allows the inheriting classes to determine the class to be created
-    const result = new this(actionType, tolerance);
+    const result = new this(ACTION_TYPE_TO_EFFECT_MODE_MAP[actionType], tolerance);
```
const-cloudinary, line 44: @strausr , how come this change does not affect other tests? like `unsharpMask`, `advancedRedEye`, etc?

const-cloudinary: if it does, maybe it's reasonable to add other tests as well?

const-cloudinary (approved):

src/actions/effect/leveled/MakeTransparent.ts
patrick-tolosa: The reason is that those variables were already handled in our internal mapping 
(See src/internal/internalConstants.ts)

This PR added `makeTransparent: 'make_transparent'`

const-cloudinary: OK

PR #519, opened 2022-02-24
Allow adding external authTokens

patrick-tolosa:
#### What does this PR solve?
This PR adds the ability to pass an external, pre-calculated authToken to the SDK.

#### Use case
The use case for this PR is using the SDK on the client side, while generating the authToken on some backend server.

## What was done in this PR?
- Add setExplicitAuthToken as a public API of the CloudinaryFile (affecting images, videos, medias etc.)
- Implement the missing `sign()` method
- Fix a bug where you could sign a URL simply by providing a signature, without calling sign() - it is now required to call sign() as intended.
- Add validations for the usage of signatures and authTokens (The SDK will throw when invalid combinations are found)
- Refactor the logic for appending query params to the URL to a new function.
- Disable analytics in case the URL contains any query param (As well as if it appears in the publicID)
- Added tests (obviously!)

## Discussion points to close before merging

- [ ] Our public function is called `sign()` while our internal flag is called `signURL` - this seems confusing from a maintenance point of view.
- [ ] We're currently URL-encoding publicIDs, turning `?` into `%3F` - this looks like a bug for public IDs that actually need the query param intact (such as for auto-upload) - I've added some code that keeps the behavior as it is today, but I'd like a second glance on this solution (This actual issue, of encoding publicIDs, will need to be addressed in a separate PR)
- [ ] I'm not content about the names `authToken` and `explicitAuthToken` , it sounds like `authTokenConfig` and `authToken` are better names - I'd love to hear opinions here.

#### Final checklist

### DIFF
--- a/__TESTS__/unit/analytics/analytics.node.test.ts
+++ b/__TESTS__/unit/analytics/analytics.node.test.ts
@@ -169,4 +169,40 @@ describe('Add analytics to a regular URL', () => {
       }
     })).toBe("https://res.cloudinary.com/demo/image/upload/sample%3Fparam%3Dtest");
   });
+
+  it('Should not include analytics when authToken is provided', () => {
+    const cldImage = createNewImageWithAnalytics('sample');
+    cldImage.sign();
+    cldImage.setExplicitAuthToken('someToken');
+    // AZ -> Algo A, SDK Code is Z
+    // Alh -> 1.24.0 from package.json
+    // AM -> 12.0.0 Underlying tech
+    // A -> responsive
+    expect(cldImage.toURL({
+      trackedAnalytics: {
+        sdkCode: 'Z',
+        sdkSemver: '1.24.0',
+        techVersion: '12.0.0',
+        responsive: true
+      }
+    })).toBe("https://res.cloudinary.com/demo/image/upload/sample?__cld_token__=someToken");
+  });
+
+  it('Should include analytics when signature is provided', () => {
+    const cldImage = createNewImageWithAnalytics('sample');
+    cldImage.sign();
+    cldImage.setSignature('someSignature');
+    // AZ -> Algo A, SDK Code is Z
+    // Alh -> 1.24.0 from package.json
+    // AM -> 12.0.0 Underlying tech
+    // A -> responsive
+    expect(cldImage.toURL({
+      trackedAnalytics: {
+        sdkCode: 'Z',
+        sdkSemver: '1.24.0',
+        techVersion: '12.0.0',
+        responsive: true
+      }
+    })).toBe(`https://res.cloudinary.com/demo/image/upload/s--someSignature--/sample?_a=AZAlhAMA`);
+  });
 });
--- a/src/assets/CloudinaryFile.ts
+++ b/src/assets/CloudinaryFile.ts
@@ -58,13 +58,15 @@ class CloudinaryFile {
   protected cloudName: string; // populated from the cloud config
   protected apiKey: string; // populated from  the cloud config
   protected apiSecret: string; // populated from the cloud config
-  protected authToken: IAuthTokenConfig; // populated from the cloud config
+  protected authToken: IAuthTokenConfig; // The configuration describing the authToken, used in a node environment to calculate the token
   protected urlConfig: IURLConfig;
 
   private version: number | string;
   private publicID: string;
   private extension: string;
   private signature: string;
+  private signURL: boolean = null; // An internal flag indicating if sign() was called,
+  private explicitAuthToken: string; // A pre-calculated authToken provided externally (is not calculated by the SDK)
   private suffix: string;
   private deliveryType: string; // type upload/private
 
@@ -86,7 +88,7 @@ class CloudinaryFile {
 
   /**
    * @description Sets the Cloud Config for this asset
-   * @param urlConfig
+   * @param cloudConfig
    * @return {this}
    */
   setCloudConfig(cloudConfig: ICloudConfig): this {
@@ -139,6 +141,14 @@ class CloudinaryFile {
   }
 
   /**
+   * Sets the __cld_token__ value for the authToken signature
+   * @param token
+   */
+  setExplicitAuthToken(token: string) {
+    this.explicitAuthToken = token;
+  }
+
+  /**
    * @description Sets the version of the asset.
    * @param {string} newVersion The version of the asset.
    * @return {this}
@@ -162,7 +172,8 @@ class CloudinaryFile {
     return this;
   }
 
-  sign(): this {
+  sign(val = true): this {
+    this.signURL = val;
     return this;
   }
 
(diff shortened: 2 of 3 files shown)


src/assets/CloudinaryFile.ts
```diff
@@ -138,6 +140,14 @@ class CloudinaryFile {
     return this;
   }
 
+  /**
+   * Sets the __cld_token__ value for the authToken signature
+   * @param token
+   */
+  setExplicitAuthToken(token: string) {
```
const-cloudinary, line 147: `setAuthToken` sounds like a better name, the word `Explicit` only introduces confusion.

src/assets/CloudinaryFile.ts
```diff
@@ -138,6 +140,14 @@ class CloudinaryFile {
...
+    this.explicitAuthToken = token;
```
const-cloudinary, line 148: I would move this to the AuthToken.
Like:
`this.AuthToken.TokenValue`  or `this.AuthToken.Token` or `this.AuthToken.Signature`

src/assets/CloudinaryFile.ts
```diff
@@ -237,14 +255,73 @@ class CloudinaryFile {
...
+    if (!this.signURL) { return; }
+
+    // Do we have an explicit authToken? if so, use it.
+    if (this.explicitAuthToken) {
+      return `__cld_token__=${this.explicitAuthToken}`;
+    }
+
+    // Do we have an authToken config? if so, use it.
+    if (this.authToken) {
+      // To be implemented, server-side calculation of the authToken.
```
const-cloudinary, line 289: I would not leave NOP code, would raise Unimplemented exception or something. Like, to make sure people get expected behavior.

const-cloudinary (changes requested):
Hey @patrick-tolosa , please see my comments.
In addition I would really like to verify whether we cannot use both `AuthToken` and `Analytics` together. It looks like it's not the use case, unless I completely miss something. `AuthToken` is parsed on the CDN level, I'm not even sure whether it is passed to the Cloudinary servers. The reason I am asking, is we get many SDKs with explicit test for `?` in public ID, and it looks like here we introduce a bug (__cld_token__ is NOT a part of the public ID). If it is so, never mind :)

const-cloudinary:
@patrick-tolosa ,
Regrading discussion points:

According to the spec, the parameter name is `signURL`, it looks like it is a bug in JS, and the name should be aligned.

I vote for  `authTokenConfig` and `authToken` names. (instead of `explicitAuthToken`)

__TESTS__/unit/analytics/analytics.node.test.ts
```diff
@@ -169,4 +169,40 @@ describe('Add analytics to a regular URL', () => {
       }
     })).toBe("https://res.cloudinary.com/demo/image/upload/sample%3Fparam%3Dtest");
   });
+
+  it('Should not include analytics when authToken is provided', () => {
+    const cldImage = createNewImageWithAnalytics('sample');
+    cldImage.sign();
```
maoznir, line 175: Why is the sign() function needed?

src/assets/CloudinaryFile.ts
patrick-tolosa: The problem here is that the word `authToken` already signifies the whole authToken config, see this: 

https://github.com/cloudinary/js-url-gen/pull/519/files/2cabdf9204ce87ad7c98d1dbafd488c09827017f#diff-bcfa6f78219d31a71565b9004ec7f94c114978d695f62b52c18dd6e759d6d9c7R61

It's a bit strange that setAuthToken here is a string, while authToken is a configuration object.
Perhaps the config name should change to `authTokenConfig`, and this one should change to `setAuthToken`?

What do you think?

src/assets/CloudinaryFile.ts
patrick-tolosa: I'm not entirely sure how I feel about this, 
Overloading the same thing can lead to confusion in runtime, but even more for the maintainers.

We can move everything to an authentication class, which can deal with everything, but we don't have that right now (And the question is do we want it?)

src/assets/CloudinaryFile.ts
```diff
@@ -237,14 +255,73 @@ class CloudinaryFile {
...
       return '';
     }
   }
 
+  /**
+   * Gets the full query param for the authToken signature
+   * - For client-side, this function uses the explicitAuthToken
+   * - For server-side, this function calculates the authToken given the authToken config
+   */
+  getAuthTokenQuery(): string {
```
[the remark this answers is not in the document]
patrick-tolosa: Keeping that in mind, I'm not sure what responsibility this class has (As my comment above) - does it contain all authenticaiton logic? or just authToken?

src/assets/CloudinaryFile.ts
patrick-tolosa: What would be the alternative? 
implementing the authToken calculation is beyond the scope of this PR, should we block this feature until it's we implement the token calculation?

src/assets/CloudinaryFile.ts
```diff
@@ -237,14 +255,73 @@ class CloudinaryFile {
...
+    }
+  }
+
+
+  /**
+   * Appends various query params to the URL encoded safe URL
+   * @param safeURL
+   * @param trackedAnalytics
+   */
+  appendQueryParamsToURL(safeURL: string, trackedAnalytics: Partial<ITrackedPropertiesThroughAnalytics>): string {
```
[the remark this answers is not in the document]
patrick-tolosa: We can(and will) simplify the appending of the query param, but quite a bit of logic will still be here. 

In any case it would probably be better, yes.

src/assets/CloudinaryFile.ts
```diff
@@ -282,14 +359,7 @@ class CloudinaryFile {
...
         .replace(/=/g, '%3D');
-
-      // urlConfig.analytics is true by default, has to be explicitly set to false to overwrite
-      // Don't add analytics when publicId includes a '?' to not risk changing existing query params
-      if (this.urlConfig.analytics !== false && !(publicID.includes('?'))) {
-        return `${safeURL}?_a=${getSDKAnalyticsSignature(trackedAnalytics)}`;
-      } else {
-        return safeURL;
-      }
+      return this.appendQueryParamsToURL(safeURL, trackedAnalytics);
```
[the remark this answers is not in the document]
patrick-tolosa: This is a bug I think we have that I didn't want to fix with this PR.

The checks are relevant because we should not be replacing `?` with `%3F`, and that should be fixed promptly. 

So yes, the checks are meaningless now, however once we fix the bug we'll need to re-add them.

__TESTS__/unit/analytics/analytics.node.test.ts
patrick-tolosa: It's a way for the user to explicitly state he wants to sign the URL, instead of us implicitly assuming he wants to because he provided a signature or a token.

patrick-tolosa:
@const-cloudinary I've updated the PR with the following key changes:

- Rename sign() to signURL()
- Rename authToken to authTokenConfig
- Add an internal class responsible for all authentication operations.
- Fixed some infra code related to compilation tests (Moving our validation function caused them to fail due to missing cloudName).

patrick-tolosa:
We'll revisit the solution in a future date

PR #527, opened 2022-05-04
Add transformation-builder package

strausr:
#### What does this PR solve?
Add transformation-builder package
Note that the package will first need to be released. We will then update the dependency here.


patrick-tolosa (commented):
I think the test are failing because cloudinaryLtd is private.
Lets move it to public and make sure the tests pass before we approve this PR.

patrick-tolosa (approved):

PR #531, opened 2022-08-08
fix: build fails on node 12

maoznir:
* Update typescript version to fix error on node 12
* Add missing parameters to npm scripts using ts-node
* Fix the export of TEST_FILE_STRUCTURE - make it named instead of default
* Fix bundle size min size comparison


rollup.config.js
```diff
@@ -41,6 +42,7 @@ export default [
       resolve(),
       replace({
         'PACKAGE_VERSION_INJECTED_DURING_BUILD': version,
+        preventAssignment: true
```
strausr, line 45: what does this do?

maoznir: There is a warning for replace plugin:
@rollup/plugin-replace: 'preventAssignment' currently defaults to false. It is recommended to set this option to true, as the next major version will default this option to true.

See here what this does: https://github.com/rollup/plugins/tree/master/packages/replace#preventassignment

strausr (approved):

PR #547, opened 2023-01-31
Re-export BackgroundRemoval and DropShadow actions

magdakwiecien:
#### What does this PR solve?
- Added suport for new actions - Background Removal and Drop Shadow

#### Final checklist


const-cloudinary (approved):

const-cloudinary:
I think, at some point we need to decouple these two packages.
The only reason we have it is the legacy compatibility.

PR #553, opened 2023-03-15
enable multiple features in analytic tokens

lbendavid-cloudinary:
#### What does this PR solve?
enable creating feature flags for more than one feature in a single call

#### Final checklist


const-cloudinary (changes requested):
Does it preserve the order how the features where invoked?
I can see from the code that you just add them in a constant order.
As you know order matters. 
It can provide useful information, for example when order is wrong or not optimal.

PR #555, opened 2023-03-29
enable passing query params to URLConfig

lbendavid-cloudinary:
#### What does this PR solve?
Add the ability to provide additional query params to be added to the final generated URL

#### Final checklist

### DIFF
--- a/__TESTS__/unit/url/url.test.ts
+++ b/__TESTS__/unit/url/url.test.ts
@@ -94,4 +94,16 @@ describe('Tests for URL configuration', () => {
 
     expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/s--${signature}--/c_crop,w_100/sample`);
   });
+
+  it('Should include query params', function () {
+    const image = createNewImage('sample', {cloudName: 'demo'}, {queryParams: {"_i": "abcde", "_z": 1234}});
+    const url = image.toURL();
+    expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_z=1234`);
+  });
+
+  it('Should include query params with analytics', function () {
+    const image = createNewImage('sample', {cloudName: 'demo'}, {analytics: true, queryParams: {"_i": "abcde"}});
+    const url = image.toURL();
+    expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_a=E`);
+  });
 });
--- a/__TESTS__/unit/url/url.test.ts
+++ b/__TESTS__/unit/url/url.test.ts
@@ -96,13 +96,13 @@ describe('Tests for URL configuration', () => {
   });
 
   it('Should include query params', function () {
-    const image = createNewImage('sample', {cloudName: 'demo'}, {urlSearchParams: {"_i": "abcde", "_z": 1234, "_t": false}});
+    const image = createNewImage('sample', {cloudName: 'demo'}, {queryParams: {"_i": "abcde", "_z": 1234, "_t": false}});
     const url = image.toURL();
     expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_z=1234&_t=false`);
   });
 
   it('Should include query params with analytics', function () {
-    const image = createNewImage('sample', {cloudName: 'demo'}, {analytics: true, urlSearchParams: {"_i": "abcde"}});
+    const image = createNewImage('sample', {cloudName: 'demo'}, {analytics: true, queryParams: {"_i": "abcde"}});
     const url = image.toURL();
     expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_a=E`);
   });
--- a/__TESTS__/unit/url/url.test.ts
+++ b/__TESTS__/unit/url/url.test.ts
@@ -103,7 +103,12 @@ describe('Tests for URL configuration', () => {
 
   it('Should include query params with analytics', function () {
     const image = createNewImage('sample', {cloudName: 'demo'}, {analytics: true, queryParams: {"_i": "abcde"}});
-    const url = image.toURL();
-    expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_a=E`);
+    const analyticsOptions = {
+      techVersion: '16.0.0',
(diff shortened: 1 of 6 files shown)


src/config/interfaces/Config/IURLConfig.ts
```diff
@@ -93,6 +94,11 @@ interface IURLConfig {
    * Whether or not to force a version
    */
   forceVersion?: boolean;
+
+  /**
+   * Additional query params to be added to the URL
+   */
+  queryParams?: Record<string, string | number>
```
pawelphilipczyk-cloudinary, line 101: I'm thinking if the name should rather be aligned with web api (URLSearchParams) so that would mean `searchParams`. We can have a discussion about it.

src/assets/CloudinaryFile.ts
```diff
@@ -283,10 +283,17 @@ class CloudinaryFile {
...
+      const queryParams: Record<string, string | number> = this.urlConfig.queryParams || {};
+
       // urlConfig.analytics is true by default, has to be explicitly set to false to overwrite
       // Don't add analytics when publicId includes a '?' to not risk changing existing query params
       if (this.urlConfig.analytics !== false && !(publicID.includes('?'))) {
-        return `${safeURL}?_a=${getSDKAnalyticsSignature(trackedAnalytics)}`;
+        queryParams._a = getSDKAnalyticsSignature(trackedAnalytics);
+      }
+
+      const queryParamsStr = `${Object.entries(queryParams).map(([key, value]) => `${key}=${value}`).join('&')}`;
```
pawelphilipczyk-cloudinary, line 294: Could be probably replaced with 
```suggestion
      const queryParamsStr = new URLSearchParams(queryParams).toString(); 
```

pawelphilipczyk-cloudinary (approved):
Looking great! 

This one will require some documentation changes right?

lbendavid-cloudinary:
The auto generated docs will include this addition, see here
https://github.com/cloudinary/js-url-gen/pull/555/files#diff-2dd82a5a4c9a71c84c66f12743e7e50339e96c54fa6593c37fa5c4a008ffbe6fR16

src/config/URLConfig.ts
```diff
@@ -13,6 +13,7 @@ class URLConfig extends Config implements IURLConfig {
   useRootPath?: boolean;
   secure?: boolean;
   forceVersion?: boolean;
+  urlSearchParams?: Record<string, string | number | boolean>;
```
const-cloudinary, line 16: Maybe just call it `queryParams` ?

src/config/URLConfig.ts
```diff
@@ -101,6 +102,14 @@ class URLConfig extends Config implements IURLConfig {
     this.forceVersion = value;
     return this;
   }
+
+  /**
+   * @param params Sets additional params
+   */
+  setUrlSearchParams(params: Record<string, string | number | boolean>):this {
```
const-cloudinary, line 109: Same here, maybe just call it `setQueryParams` ?

const-cloudinary (changes requested):

src/config/URLConfig.ts
lbendavid-cloudinary: actually it was `queryParams` but we wanted to be aligned wit web api
https://github.com/cloudinary/js-url-gen/pull/555#discussion_r1151699317

const-cloudinary: @lbendavid-cloudinary, @pawelphilipczyk-cloudinary 
Is Web API very popular? I hear this name for the first time. And we will have to align this implementation with other SDKs(at some point). In all other languages it is called query parameters/query string.

const-cloudinary: The class is called URLConfig, having `url` prefix in `urlSearchParams` looks redundant, no matter what name we choose, it should be without `url` prefix.

pawelphilipczyk-cloudinary: @const-cloudinary It's not about popularity, `URLSearchParams` is the standard name for the [interface in Web API](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams). People call it query params and I did too until learning this. 

If we want to align with other SDKs then I take that as a fair point and back down with my initial comments.

const-cloudinary: The first line you read in the description is:
`defines utility methods to work with the query string of a URL.`
And in RFC the component is called "Query":
https://www.rfc-editor.org/rfc/rfc3986#section-3.4

But if you insist, I'm OK with that.

lbendavid-cloudinary: I think having a consistent naming across our SDK's is important, also from docs point of view. So i think `queryParams` would be the better name

src/config/URLConfig.ts
```diff
@@ -13,6 +13,7 @@ class URLConfig extends Config implements IURLConfig {
   useRootPath?: boolean;
   secure?: boolean;
   forceVersion?: boolean;
+  queryParams?: Record<string, string | number | boolean>;
```
const-cloudinary, line 16: Can query params value be an array or dict?

__TESTS__/unit/url/url.test.ts
```diff
@@ -94,4 +94,16 @@ describe('Tests for URL configuration', () => {
...
+  it('Should include query params', function () {
+    const image = createNewImage('sample', {cloudName: 'demo'}, {queryParams: {"_i": "abcde", "_z": 1234, "_t": false}});
+    const url = image.toURL();
+    expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_z=1234&_t=false`);
+  });
+
+  it('Should include query params with analytics', function () {
+    const image = createNewImage('sample', {cloudName: 'demo'}, {analytics: true, queryParams: {"_i": "abcde"}});
+    const url = image.toURL();
+    expect(url).toEqual(`https://res.cloudinary.com/demo/image/upload/sample?_i=abcde&_a=E`);
```
const-cloudinary, line 107: `_a=E` is not the best test (`E` stands for error) it would be better to have a valid/mocked analytics token.

const-cloudinary (changes requested):

__TESTS__/unit/url/url.test.ts
lbendavid-cloudinary: thanks, fixed

const-cloudinary (approved):

PR #563, opened 2023-04-24
change algo letter in analytics

lbendavid-cloudinary:
#### What does this PR solve?
Add "product" field to analytics token

#### Final checklist

### DIFF
--- a/__TESTS__/unit/analytics/analytics.node.test.ts
+++ b/__TESTS__/unit/analytics/analytics.node.test.ts
@@ -43,7 +43,7 @@ describe('Add analytics to a regular URL', () => {
         techVersion: '12.0.0',
         accessibility: true
       }
-    })).toContain('?_a=AZAlhAMD');
+    })).toContain('?_a=BAZAlhAMD');
   });
 
   it('Test lazyload feature value', () => {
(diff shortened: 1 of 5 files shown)


__TESTS__/unit/analytics/analytics.node.test.ts
```diff
@@ -43,7 +43,7 @@ describe('Add analytics to a regular URL', () => {
         techVersion: '12.0.0',
         accessibility: true
       }
-    })).toContain('?_a=AZAlhAMD');
+    })).toContain('?_a=BAZAlhAMD');
```
pawelphilipczyk-cloudinary, line 46: Shouldn't this be `BZAlhAMD` (only first letter changed from A to B?)

lbendavid-cloudinary: Algorithm "B" also introduce a new letter fro product, which has a default of "A",
so as a result, "AZ..." will be "BAZ..."

const-cloudinary (approved):

const-cloudinary (changes requested):
Please fix travis build. There is a failing test.

const-cloudinary (approved):
The PR is good.
Please make sure to write a good commit message when you merge it.
e.g. something like: 
`Update analytics token to algorithm B`

PR #578, opened 2023-10-16
change analytics to algorithm C

lbendavid-cloudinary:
#### What does this PR solve?
Implement algorithm C (OS type and version)

#### Final checklist

### DIFF
--- a/src/sdkAnalytics/getSDKAnalyticsSignature.ts
+++ b/src/sdkAnalytics/getSDKAnalyticsSignature.ts
@@ -31,6 +31,8 @@ function ensureShapeOfTrackedProperties(trackedAnalytics?: Partial<ITrackedPrope
     sdkCode: 'T', // Base code
     sdkSemver : packageVersion.split('-')[0], // remove -beta, -alpha or other tagged versions from the version string
     product: 'A',
+    osType: 'Z',
+    osVersion: '0.0',
     responsive: false,
     placeholder: false,
     lazyload: false,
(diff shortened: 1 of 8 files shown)


src/sdkAnalytics/getSDKAnalyticsSignature.ts
```diff
@@ -31,6 +31,8 @@ function ensureShapeOfTrackedProperties(trackedAnalytics?: Partial<ITrackedPrope
     sdkCode: 'T', // Base code
     sdkSemver : packageVersion.split('-')[0], // remove -beta, -alpha or other tagged versions from the version string
     product: 'A',
+    osType: 'Z',
```
const-cloudinary, line 34: Don't we want to actually detect it?

const-cloudinary (changes requested):
I find it useful to detect actual values for os type and version.

src/sdkAnalytics/getSDKAnalyticsSignature.ts
adimiz1: I think it'll be dependent on the inheriting  framework
For example, React Native will be the one to detect the platform and the os version.

adimiz1 (approved):

src/sdkAnalytics/getSDKAnalyticsSignature.ts
const-cloudinary: This package can be used as a standalone package in your pure JS project. We can detect os type and version.

PR #581, opened 2023-10-31
fix: double encoding of commas in createCloudinaryURL

### LINKED ISSUE #580: Double encoding of comma in createCloudinaryURL()
In the createCloudinaryURL method, the comma , character in the publicID is being encoded twice.

This results in %2C being transformed into %252C, which causes issues when trying to access images with a comma in their publicID.

This issue is causing 404 errors when trying to access images with a comma in their publicID, as the double-encoded comma does not match the actual publicID of the image.

https://github.com/cloudinary/js-url-gen/blob/master/src/assets/CloudinaryFile.ts

the issue is in these lines of code

```js
const publicID = this.publicID
  .replace(/,/g, '%2C');

const safeURL = encodeURI(url)
  .replace(/\?/g, '%3F')
  .replace(/=/g, '%3D')
```

A fix could be to replace the double encoded characters so single encoded

```js
const safeURL = encodeURI(url)
  .replace(/\?/g, '%3F')
  .replace(/=/g, '%3D')
  .replace(/%252C/g, '%2C');
```

I've tested this fix in my application(s) with multiple characters, commas, and slashes, and it seems to work correctly.

Edjevw12:
fixes #580

#### What does this PR solve?
double encoding of commas

#### Final checklist

### DIFF
--- a/src/assets/CloudinaryFile.ts
+++ b/src/assets/CloudinaryFile.ts
@@ -281,7 +281,9 @@ class CloudinaryFile {
     } else {
       const safeURL = encodeURI(url)
         .replace(/\?/g, '%3F')
-        .replace(/=/g, '%3D');
+        .replace(/=/g, '%3D')
+        // Replace double-encoded comma with single-encoded comma to prevent issues with publicIDs containing commas
+        .replace(/%252C/g, '%2C');
 
       const queryParams = new URLSearchParams(this.urlConfig.queryParams as Record<string, string>);
 


Edjevw12:
```js
 Tests for Encoding the URL › Should encode cloudinary characters (",") in a publicID
    expect(received).toBe(expected) // Object.is equality
    Expected: "https://res.cloudinary.com/demo/image/upload/sam%252Cple"
    Received: "https://res.cloudinary.com/demo/image/upload/sam%2Cple"
      20 |     const url = createNewImage('sam,ple')
      21 |       .toURL();
    > 22 |     expect(url).toBe('https://res.cloudinary.com/demo/image/upload/sam%252Cple');
         |                 ^
      23 |   });
      24 |
      ```

Seems like this is "expected" behaviour but causes 404s?

image

